PR #23, opened 2019-10-09
Make error handling more resilient:

MalloZup:
Don't panic on error just log them and skip iteration with continue

### DIFF
--- a/ha_cluster_exporter.go
+++ b/ha_cluster_exporter.go
@@ -14,6 +14,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+// this types are for reading pacemaker configuration xml when running crm_mon command
+// and lookup the corrispective value
 type crmMon struct {
 	Version string  `xml:"version,attr"`
 	Summary summary `xml:"summary"`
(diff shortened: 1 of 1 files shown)


ha_cluster_exporter.go
```diff
@@ -149,7 +151,7 @@ func resetClusterMetrics() {
 	err := prometheus.Register(nodeResources)
 	if err != nil {
 		log.Println("[ERROR]: failed to register NodeResource metric. Perhaps another exporter is already running?")
-		log.Panic(err)
+		log.Println(err)
 	}
```
stefanotorresi: maybe we should return this error instead, and use the logger in the caller (`main()`)?

ha_cluster_exporter.go
```diff
@@ -163,7 +165,7 @@ func resetClusterMetrics() {
 	err = prometheus.Register(clusterNodes)
 	if err != nil {
 		log.Println("[ERROR]: failed to register clusterNode metric. Perhaps another exporter is already running?")
-		log.Panic(err)
+		log.Println(err)
 	}
```
stefanotorresi: maybe we should return this error instead, and use the logger in the caller (`main()`)?

ha_cluster_exporter.go
```diff
@@ -218,7 +221,8 @@ func main() {
 			// todo: reflect if we want to error out. We could just ignore error for resiliance
```
stefanotorresi: I think we can remove this comment now

MalloZup:
Removing comment yep good catch 'I will check if we can return in main error. Since we are continuing the for loop so returning will break it. I will check tomorrow thx for review

PR #26, opened 2019-10-11
Implement DRBD metrics for disk-state

MalloZup:
# Description

This pr will implement this metric:

```

# HELP ha_cluster_drbd_resource_disk_state show per resource name, its role, the volume and disk_state (Diskless,Attaching, Failed, Negotiating, Inconsistent, Outdated, DUnknown, Consistent, UpToDate)
# TYPE ha_cluster_drbd_resource_disk_state gauge
ha_cluster_drbd_resource{disk_state="uptodate",resource_name="1-single-0",role="Secondary",volume="0"} 1
ha_cluster_drbd_resource{disk_state="uptodate",resource_name="1-single-1",role="Secondary",volume="0"} 1
ha_cluster_drbd_resource{disk_state="uptodate",resource_name="vg1",role="Secondary",volume="0"} 1
ha_cluster_drbd_resource{disk_state="uptodate",resource_name="vg2",role="Secondary",volume="0"} 1
ha_cluster_drbd_resource{disk_state="uptodate",resource_name="vg3",role="Secondary",volume="0"} 1
```

# Usage:

- Check if the state of DRBD disk are UptoDate, out of sync etc/
- monitor or alert drbd resource disks

## what is it is missing:

- [x] parse and populated types. 
- [x] Tests 
- [x] set the prometheus exporter metric according to metric

### DIFF
--- a/drbd_metrics.go
+++ b/drbd_metrics.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"log"
 	"os/exec"
 )
@@ -21,4 +22,13 @@ func getDrbdInfo() []byte {
 	log.Println("[INFO]: Reading drbd status with drbdsetup status ...")
 	drbdStatusRaw, _ := exec.Command("/sbin/drbdsetup", "status", "--json").Output()
 	return drbdStatusRaw
+}
+
+func parseDrbdStatus(statusRaw []byte) ([]drbdStatus, error) {
+	var drbdDevs []drbdStatus
+	err := json.Unmarshal(statusRaw, &drbdDevs)
+	if err != nil {
+		return drbdDevs, err
+	}
+	return drbdDevs, nil
 }
--- a/drbd_metrics.go
+++ b/drbd_metrics.go
@@ -10,7 +10,8 @@ type drbdStatus struct {
 	Name    string `json:"name"`
 	Role    string `json:"role"`
 	Devices []struct {
-		Volume int `json:"volume"`
+		Volume    int    `json:"volume"`
+		DiskState string `json:"disk-state"`
 	} `json:"devices"`
 }
 
--- a/drbd_metrics.go
+++ b/drbd_metrics.go
@@ -1,18 +1,14 @@
 package main
 
 import (
-	"fmt"
-	"io/ioutil"
-	"os"
+	"log"
 	"os/exec"
-	"regexp"
-	"strings"
 )
 
 // return drbd status in byte raw json
 func getDrbdInfo() []byte {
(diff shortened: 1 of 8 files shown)


stefanotorresi:
Just a nitpicky remark about naming (yeah, I might be kind of a naming freak):
The `_state` part in `ha_cluster_drbd_resource_disk_state` feels somewhat redundant. I mean, do we track anything other than the disks state?
Also, are there other disks other than "resource disks"? If not, could this be just `ha_cluster_drbd_disk`?
In general, what other `ha_cluster_drbd_*` metrics we have in our future plans? So that we can plan their naming accordingly.

MalloZup:
Good point. I think we can remove the last 2 words

drbd_metrics.go
```diff
@@ -0,0 +1,34 @@
...
+		Volume    int    `json:"volume"`
+		DiskState string `json:"disk-state"`
+	} `json:"devices"`
+}
+
+// return drbd status in byte raw json
+func getDrbdInfo() []byte {
+	// get ringStatus
+	log.Println("[INFO]: Reading drbd status with drbdsetup status ...")
+	drbdStatusRaw, _ := exec.Command("/sbin/drbdsetup", "status", "--json").Output()
```
stefanotorresi: we could return this error and log it in the main, I'm not super comfortable in suppressing it completely; what do you think?

ha_cluster_exporter.go
```diff
@@ -189,15 +183,23 @@ func main() {
...
 			if err != nil {
 				log.Println(err)
+				time.Sleep(time.Duration(int64(*timeoutSeconds)) * time.Second)
 				continue
 			}
 			// retrieve a list of sbd devices
-			sbdDevices := getSbdDevices(sbdConfiguration)
+			sbdDevices, err2 := getSbdDevices(sbdConfiguration)
+			// mostly, the sbd_device were not set in conf file for returning an error
+			if err2 != nil {
```
stefanotorresi: isn't common idiom in Go to just mask always the same `err` variable?

MalloZup: yop

MalloZup: I think i did atypo somewhere :grin:

drbd_metrics.go
MalloZup: ok

MalloZup:
@storresi can web merge?

MalloZup:
OK THX!  I have added it to the release draft. We will wait a bit this time before releasing a new rpm since we need to do some refactoring etc

Issue #28, opened 2019-10-11
Bug / Feature Request: Cluster SBD Error'ing (when not present)


ashleyprimo:
I have noticed that when operating the exporter on a machine that does not have `sbd` setup the following set of errors are seen repeatedly:

```
2019/10/11 15:14:12 [INFO]: Reading cluster SBD configuration..
2019/10/11 15:14:12 [ERROR] Could not open sbd config file open /etc/sysconfig/sbd: no such file or directory
```

There should maybe be a check such as 

```
if _, err := os.Stat("/etc/sysconfig/sbd"); os.IsNotExist(err) {
	log.Println("[INFO]: SBD is not present... skipping.")
} else {

}
```

This would give a single output of `0000/00/00 00:00:00 [INFO]: SBD is not present... skipping.` instead of continually repeating an error in the absence of sbd. :)

MalloZup:
done

MalloZup:
thx!

PR #29, opened 2019-10-11
Regarding Issue #28 - Fix SBD error in absence of SBD

ashleyprimo:
This is in relation #28 and proposes a quick fix to resolve errors seen when `/etc/sysconfig/sbd` is not present on the machine.

### DIFF
--- a/ha_cluster_exporter.go
+++ b/ha_cluster_exporter.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"log"
 	"net/http"
+	"os"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -183,6 +184,10 @@ func main() {
 
 	// set SBD device metrics
 	go func() {
+		if _, err := os.Stat("/etc/sysconfig/sbd"); os.IsNotExist(err) {
+			return
+		}
+
 		for {
 			log.Println("[INFO]: Reading cluster SBD configuration..")
 			// read configuration of SBD


ha_cluster_exporter.go
```diff
@@ -183,34 +184,38 @@ func main() {
...
-			// retrieve a list of sbd devices
-			sbdDevices := getSbdDevices(sbdConfiguration)
-			// set and return a map of sbd devices with true healthy, false not
-			sbdStatus := setSbdDeviceHealth(sbdDevices)
-
-			if len(sbdStatus) == 0 {
-				log.Println("[WARN]: Could not retrieve any sbd device")
-				continue
-			}
+		if _, err := os.Stat("/etc/sysconfig/sbd"); os.IsNotExist(err) {
```
MalloZup: I like the check. The only thing which I would try to do is that if we could avodid if/else and just use guard clause.

e.g:

 _, err := os.Stat("/etc/sysconfig/sbd")
// check if file exist
if os.IsExist(err) {

// do sbd metrics
}


withing this we don't log however that we skip the `sbd` metric but in term of cognition overload is more simple to read

ashleyprimo: That makes sense, pushed the suggested change. :slightly_smiling_face:

MalloZup:
@ashleyprimo please squash commits into 1 and create a minimal good commit msg. A part of this great catch :100:

ha_cluster_exporter.go
```diff
@@ -183,34 +184,36 @@ func main() {
...
+						sbdDevStatus.WithLabelValues(sbdDev).Set(float64(1))
+					} else {
+						sbdDevStatus.WithLabelValues(sbdDev).Set(float64(0))
+					}
 				}
-			}
 
-			time.Sleep(time.Duration(int64(*timeoutSeconds)) * time.Second)
+				time.Sleep(time.Duration(int64(*timeoutSeconds)) * time.Second)
+			}
```
stefanotorresi: rather than indenting the whole `for` loop, can I suggest the "fail fast, fail early" approach? I.e. just add an if with inverted conditional logic and return out of the goroutine early if the condition is not met.

ashleyprimo: Made this change, :+1:

MalloZup: yes good catch! That is a guard clause approach which make more sense

MalloZup: did you tested manually for any case? should look good to me anyways thx ! :sun_with_face:

stefanotorresi: @MalloZup I'll add some unit tests in my refactoring.

Issue #46, opened 2019-10-15
Install logrous in golang modules


MalloZup:
# Desc:

we should use logrous (https://github.com/sirupsen/logrus) for a more fine grained control over logs.

# Tasks:

- install logrous in go module file

# Not mandatory
- 2) bonus convert some logs info to debug mode

stefanotorresi:
implemented by #48

PR #48, opened 2019-10-15
Various refactorings

stefanotorresi:
This PR applies various extensive refactorings to consolidate the codebase:

- apply some guidelines from https://github.com/golang/go/wiki/CodeReviewComments;
- remove output from any internal function and move logging to the application outer layer, the `main()`, applying error wrapping instead where applicable;
- apply the `ha_cluster_` prefix to all the metric names;
- rename the `node` metric attribute to `node_name`, to make it consistent with other attributes like `resource_name` and the upcoming `cluster_name`;
- implements #46;
- remove the deployment script and the corresponding Makefile target, since it's a single `scp` command that can be onelined easily by users and developers alike.


pacemaker_metrics.go
```diff
@@ -2,6 +2,8 @@ package main
 
 import (
 	"encoding/xml"
+
+	"github.com/pkg/errors"
```
MalloZup: I think we can spare us this external dep

and using `return fmt.Errorf("Error serializing libvirt storage pool: %s", err)`

ha_cluster_exporter.go
```diff
@@ -258,11 +258,11 @@ func main() {
 				started = true
 			}
 
-			log.Println("[INFO]: Reading ring status...")
+			log.Infoln("Reading ring status...")
```
MalloZup: @stefanotorresi  the INFO here I think it might be more debug thing . see the issue we had. I'm ok if we consider this `info` -> debug for later.

ha_cluster_exporter.go
```diff
@@ -282,11 +282,11 @@ func main() {
 				started = true
 			}
 
-			log.Println("[INFO]: Reading quorum status...")
+			log.Infoln("Reading quorum status...")
```
MalloZup: ditto. All this info thing it might be only debug. I think for moment we can just fix it later

Makefile
```diff
@@ -11,15 +11,12 @@ vet-check:
 
 fmt-check:
 	go fmt  .
+
 test:
-	go test 
+	go test -v
```
MalloZup: +1

PR #51, opened 2019-10-16
add remote disk metric

MalloZup:
# desc:

1) create a metric like

```
# HELP ha_cluster_drbd_resources_remote_connection show per remote connection resource name, its role, the volume and disk_state (Diskless,Attaching, Failed, Negotiating, Inconsistent, Outdated, DUnknown, Consistent, UpToDate)
# TYPE ha_cluster_drbd_resources_remote_connection gauge
ha_cluster_drbd_resources_remote_connection{peer_disk_state="dunknown",peer_node_id="0",peer_role="Unknown",resource_name="drbd_passive",volume="0"} 1
```

2) fix some error handling things:
 - use warn instead of error
 - other minor refactor and improvements

### DIFF
--- a/ha_cluster_exporter.go
+++ b/ha_cluster_exporter.go
@@ -74,10 +74,10 @@ var (
 			Help: "show per resource name, its role, the volume and disk_state (Diskless,Attaching, Failed, Negotiating, Inconsistent, Outdated, DUnknown, Consistent, UpToDate)",
 		}, []string{"resource_name", "role", "volume", "disk_state"})
 
-	drbdRemoteDiskState = prometheus.NewGaugeVec(
+	remoteDrbdDiskState = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "ha_cluster_drbd_resources_remote_connection",
-			Help: "show per remote connection resource name, its role, the volume and disk_state (Diskless,Attaching, Failed, Negotiating, Inconsistent, Outdated, DUnknown, Consistent, UpToDate)",
+			Help: "show per remote connection resource name, its role, peer-id , the volume and disk_state (Diskless,Attaching, Failed, Negotiating, Inconsistent, Outdated, DUnknown, Consistent, UpToDate)",
 		}, []string{"resource_name", "peer_node_id", "peer_role", "volume", "peer_disk_state"})
 )
 
@@ -91,7 +91,7 @@ func init() {
 	prometheus.MustRegister(corosyncQuorate)
 	prometheus.MustRegister(sbdDevStatus)
 	prometheus.MustRegister(drbdDiskState)
-	prometheus.MustRegister(drbdRemoteDiskState)
+	prometheus.MustRegister(remoteDrbdDiskState)
 }
 
 // this function is for some cluster metrics which have resource as labels.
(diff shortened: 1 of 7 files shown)


ha_cluster_exporter.go
```diff
@@ -149,50 +165,60 @@ func main() {
...
 		log.Infoln("Starting DRBD metrics collector...")
-
-		started := false
-
 		for {
-			if started {
-				time.Sleep(time.Duration(int64(*timeoutSeconds)) * time.Second)
-			} else {
-				started = true
-			}
```
stefanotorresi: This was here to avoid sleeping on the first iteration, since users can set the timeout arbitrarily. You didn't like it? :crying_cat_face:

ha_cluster_exporter.go
```diff
@@ -134,7 +141,16 @@ func resetDrbdMetrics() error {
 	if err != nil {
 		return errors.Wrap(err, "failed to register DRBD disk state metric. Perhaps another exporter is already running?")
 	}
+	prometheus.Unregister(drbdRemoteDiskState)
+	drbdRemoteDiskState = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "ha_cluster_drbd_resources_remote_connection",
+			Help: "show per remote connection resource name, its role, the volume and disk_state (Diskless,Attaching, Failed, Negotiating, Inconsistent, Outdated, DUnknown, Consistent, UpToDate)",
+		}, []string{"resource_name", "peer_node_id", "peer_role", "volume", "peer_disk_state"})
```
stefanotorresi: Is it normal that this metric re-registration has `peer_node_id` while the first one in line 81 hasn't?
BTW, we should investigate a way to avoid the duplication in all the resets functions, but that's for another PR

ha_cluster_exporter.go
MalloZup: @stefanotorresi  nope. Imho we are complictating the thing. for the 1st iteration, most of the user will even not notice that it sleeps 5 sec

ha_cluster_exporter.go
MalloZup: @stefanotorresi  yes it is normal. peer_node_id is something more on remote concept of SBD disk. the 1st metric is only regarding the local sbd disk. The 2nd metric we need an ide because it could be multiple host having same volume

MalloZup: I agree that the reset metric mechanism could be improved. At that time I couldn't find anything convenient.

ha_cluster_exporter.go
stefanotorresi: I see your point, OTOH as a user I could set the period to 30 seconds, and nothing would happen until that time passes at least once, and to be honest I wouldn't like this UX at all...

MalloZup:
@stefanotorresi  feel free to review PR https://github.com/ClusterLabs/ha_cluster_exporter/pull/53

I have split the refactor. I'm labeling this PR as blocked because we need a DRBD env. for testing it. I can rebase it then on tthe latest change when time it will come

ha_cluster_exporter.go
stefanotorresi: I'm sorry, man, but we'll have to agree to disagree on this one...  :stuck_out_tongue_winking_eye: 

If you proposed a different, more elegant solution, I would be totally fine with it, but rather than just removing the functionality, I think we should leave it as it is, instead of knowingly introducing a bug for the sake of subjective readability.

Maybe we should ask for another opinion?

MalloZup: I don't introduce any bug , is just a different behavior. Your example with 30 Seconds is more imho rather unrealistic ( we need to give real-time metrics) so I would expect 10 Seconds as maximal timeout period. So waiting 10 second as first start is not really an issue. E.g we have software that initialize with much more initial timeout and user are ok. 

A part of the IF and else cognitive load, there is also the wrong variable name `started` doesn't refer much to anything. ( this should improved).

I didn't catch this on the previous PR because was to much things changed. :grin: 

@ayoub-belarbi thoughts on this?

MalloZup: the variable `started` if we decided to keep it, we should rename and add some comments to explain its mechanism. Started is to much ambigous

MalloZup: @ashleyprimo  any pov here?

MalloZup:
let me try with a defer approach

ha_cluster_exporter.go
MalloZup: ok fixed

MalloZup:
thx

Issue #55, opened 2019-10-16
Bug: DRBD Error'ing (when not present)


ashleyprimo:
Similar to #28 - an error is seen when DRBD is not set up on the system,
```
2019/10/16 14:34:28 fork/exec /sbin/drbdsetup: no such file or directory
```

ashleyprimo:
Will create PR shortly

MalloZup:
@ashleyprimo  thx for this. I'm ok. Anyways the erroring out is fixed by last PR in Review.

But I'm ok to add a check on the begin of the golang routine to improve the error msg/UX

ashleyprimo:
I must've missed this, just looked our CI refused to build the versions due to `go mod tidy` of the `go.mod` - internal issue, i'll fix that :slightly_smiling_face:

PR #56, opened 2019-10-16
DRBD Go Routine Improvement 

ashleyprimo:
Regarding Issue #55

### DIFF
--- a/ha_cluster_exporter.go
+++ b/ha_cluster_exporter.go
@@ -151,6 +151,11 @@ func main() {
 	// set DRBD metrics
 
 	go func() {
+		if _, err := os.Stat("/sbin/drbdsetup"); os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


ha_cluster_exporter.go
```diff
@@ -151,6 +151,11 @@ func main() {
 	// set DRBD metrics
 
 	go func() {
+        if _, err := os.Stat("/sbin/drbdsetup"); os.IsNotExist(err) {
+                log.Warnln("drbdsetup binary not available, DRBD metrics won't be collected")
+                return
+        }
+
```
stefanotorresi: the indentation is off, could you fix it? Also in Go tabs are used instead of spaces!

ashleyprimo: Yep, I use tabs as standard - not sure what happened there, but it's fixed! :+1:

MalloZup:
we might add go fmt for this cosmetical stuff lint check

PR #66, opened 2019-10-18
RFC: redesign to use `prometheus.Collector` API

stefanotorresi:
This PR proposes a substantial redesign of the project by using the `prometheus.Collector` as documented here:
- https://godoc.org/github.com/prometheus/client_golang/prometheus#hdr-Custom_Collectors_and_constant_Metrics
- https://godoc.org/github.com/prometheus/client_golang/prometheus#Collector

### Disclaimer
This restructuring is quite pervasive, and I realise that it does not correspond to the original vision by @MalloZup, who is the original author of this project, so make of this what you will! Although it took a lot of coding, this is still just a proposal, and I will accept any kind of feedback.
I originally said I would provide just a small POC for a few metrics, but then I realised that it made more sense to provide a full refactoring in the PR, because coexisting old and new code would generate too much confusion.

### Rationale

The main shortcoming of the current design is that we are using mutable structs for the metrics, which are not intended for usage in our scenario, where the source of truth is external to the exporter itself. 

As documented in the above links, in fact, we don't need the full interfaces of these metric types like `Gauge` or `GaugeVec`. These mutable metrics are meant for in-application instrumentation, where the structs persist and are mutated within the life-cycle of the application that is being monitored.
We, instead, only need to generate the metrics once per request, output them via the HTTP handler, then throw them away.

We actually currently do this in an arguably quirky way: by de-registering and registering the same metrics over and over again in a persistent loop. This involves a lot of code repetition and makes the codebase a bit harder to understand and maintain, IMHO.
But it's not required at all: we don't even need separate goroutines for the collection process, because [Prometheus already does this for us](https://github.com/prometheus/client_golang/blob/55450579111f95e3722cb93dec62fe9e847d6130/prometheus/registry.go#L439-L454).

By implementing a number of `prometheus.Collector` interfaces, [we just need to register them and launch the handler](https://github.com/stefanotorresi/ha_cluster_exporter/blob/63bad1d33193b1acc23ab82be08b8270efbbc6d1/ha_cluster_exporter.go#L59-L94). What before was the body of an iterating loop inside a goroutine, is now the body of the `Collect` method (e.g. [1] vs [2]). That's it. The core logic is pretty much the same, but now we have a clearer separation of concerns and less things to worry about.

[1]: https://github.com/ClusterLabs/ha_cluster_exporter/blob/6f568dbb59afa0e2ca7a2c35d01163fd89993eca/ha_cluster_exporter.go#L283-L338
[2]: https://github.com/stefanotorresi/ha_cluster_exporter/blob/63bad1d33193b1acc23ab82be08b8270efbbc6d1/corosync_metrics.go#L53-L88

The metrics are now declared with the `Prometheus.Desc` struct, and then created on the fly with `prometheus.NewConstMetric` every time a request comes in. We don't even need to loop anymore, because all these metrics will be provided in realtime (the frequency of update now is correctly off-loaded to downstream, i.e. [Prometheus scraping config](https://github.com/SUSE/ha-sap-terraform-deployments/blob/0ec2bb6960cd0490c299406d83c6eed8ff8f604f/salt/monitoring/prometheus/prometheus.yml#L2-L5)).  
Also, since they are not persistent, if - at the time of request - a specific metric line doesn't exist, it just won't be generated, so we don't need to de-register anything.  
We also serialize the collection process by using a mutex in all the Collectors, as exemplified in the [haproxy_exporter](https://github.com/prometheus/haproxy_exporter/blob/0ddc4bc5cb4074ba95d57257f63ab82ab451a45b/haproxy_exporter.go#L261-L263), which I took as the main reference for this proposal.

### ~~New user features~~

~~The PR also adds:~~
- ~~the new `ha_cluster_pacemaker_stonith_enabled` metric, exemplified [here](https://github.com/stefanotorresi/ha_cluster_exporter/blob/63bad1d33193b1acc23ab82be08b8270efbbc6d1/test/pacemaker.metrics#L22-L24)~~
- ~~the `timestamp` field to all the metrics, as specified by the [metrics line protocol](https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md#text-format-details).~~

edit: new features moved to separate PRs for a cleaner diff

### Other changes

The PR is also massive because when I start refactoring I always get carried away... :grin: 

The hereby presented POC hence includes various other improvements, like the introduction of `prometheus/testutil` to test the actual output of the `Collect` method in every collector, and the provisioning of all the mocks for the various external tools that we invoke via `os/exec`.

The timestamping feature also required to provide a way to mock `time.Now()`, and the `init()` method was removed because it would be also executed in the tests, which was not desirable.

I also took the chance to rename a few things by leveraging some concepts provided by the Prometheus SDK like _namespace_ and _subsystems_, which can be easily interpolated with the `prometheus.BuildFQName` method, but these new names are of course up to debate.


corosync_metrics.go
```diff
@@ -48,41 +119,42 @@ func parseQuoromStatus(quoromStatus []byte) (map[string]int, string, error) {
...
 	quorum, _ := strconv.Atoi(numberOnly.FindString(strings.SplitAfterN(quoromRaw, "Quorum:", 2)[1]))
 
-	voteQuorumInfo := map[string]int{
-		"expectedVotes":   expVotes,
-		"highestExpected": highVotes,
-		"totalVotes":      totalVotes,
-		"quorum":          quorum,
+	quorumVotes = map[string]int{
+		"expected_votes":   expVotes,
+		"highest_expected": highVotes,
```
MalloZup: I'm not sure if its golang convention to name keys of maps `foo_bar`.Afaik golang use `fooBar`

test/fake_drbdsetup.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
```
MalloZup: not sure if providing `fixture` is better then before. 
I was actually thinking to create data in file, but honestly imho it was better in the test itself so we have less data.

Btw why is this not json ?

corosync_metrics.go
```diff
@@ -7,24 +7,95 @@ import (
...
+var (
+	corosyncMetrics = metricDescriptors{
+		// the map key will function as an identifier of the metric throughout the rest of the code;
+		// it is arbitrary, but by convention we use the actual metric name
+		"quorate":           NewMetricDesc("corosync", "quorate", "Whether or not the cluster is quorate", nil),
+		"ring_errors_total": NewMetricDesc("corosync", "ring_errors_total", "Total number of corosync ring errors", nil),
+		"quorum_votes":      NewMetricDesc("corosync", "quorum_votes", "Cluster quorum votes; one line per type", []string{"type"}),
+	}
+
+	corosyncTools = map[string]string{
```
MalloZup: a map for 2 string is imho really overkill here

corosync_metrics.go
MalloZup: is cosmetical in both direction but I would rather follow golang  convention of namings

test/sbd.metrics
```diff
@@ -0,0 +1,4 @@
+# HELP ha_cluster_sbd_device_status Health status for each SBD device; 1 is healthy, 0 is not
+# TYPE ha_cluster_sbd_device_status gauge
+ha_cluster_sbd_device_status{device_name="/dev/vdc"} 1 1
```
MalloZup: why is there `1 1` here?

corosync_metrics_test.go
```diff
@@ -30,30 +30,27 @@ func TestQuoromMetricParsing(t *testing.T) {
 		Nodeid      Votes Name
 	1084780051          1 dma-dog-hana01 (local)
 	1084780052          1 dma-dog-hana02
-	dma-dog-hana01:~ # 
 	`
-	getQuoromClusterInfo()
 	voteQuorumInfo, quorate, _ := parseQuoromStatus([]byte(quoromStatus))
 
-	if voteQuorumInfo["expectedVotes"] != 232 {
+	if voteQuorumInfo["expected_votes"] != 232 {
```
MalloZup: here Dunno if `foo_bar` is good in golang hashes sinnce the var naming convention is `fooBar`

corosync_metrics.go
stefanotorresi: These keys maps 1:1 to the metric labels, so these are snake_cased to follow the Prometheus convention.
Having said that, I couldn't find any convention about map keys in the [Go style guide](https://github.com/golang/go/wiki/CodeReviewComments).

test/sbd.metrics
stefanotorresi: The latter 1 is the timestamp in millisecond. All tests use the same timestamp, initialized here: https://github.com/ClusterLabs/ha_cluster_exporter/blob/81da50bf06cd9e5e3502f0c201226099151c6122/ha_cluster_exporter_test.go#L15-L19

corosync_metrics_test.go
stefanotorresi: same as https://github.com/ClusterLabs/ha_cluster_exporter/pull/66#discussion_r336948397

corosync_metrics.go
MalloZup: yep let's ignore it for moment I'm ok

corosync_metrics_test.go
MalloZup: yeo

MalloZup:
@stefanotorresi to me looks good feel free to clean commit message and drop the extra features. so I can take a final review.

Additional note:

- we should remove doc/design.md and its links since are depracted by this.

- we could perhaps some documentation for devs for creating new metrics or collect some minimal pointer which can help to retrieve such info etc for replacing the old design.md doc.

what do you think?

stefanotorresi:
> * we should remove doc/design.md and its links since are depracted by this.
> * we could perhaps some documentation for devs for creating new metrics or collect some minimal pointer which can help to retrieve such info etc for replacing the old design.md doc.

Right, most definitely! I will update the docs accordingly.

test/sbd.metrics
stefanotorresi: I will make this more clear with a different value, I now realise that `1` is not very expressive.

stefanotorresi:
@MalloZup documentation is now updated, ready for a final review.

here are the most relevant files:
 - https://github.com/ClusterLabs/ha_cluster_exporter/blob/debded2c1335ea777cab65d2a9547d551c7308b5/doc/design.md
- https://github.com/ClusterLabs/ha_cluster_exporter/blob/debded2c1335ea777cab65d2a9547d551c7308b5/doc/metric_spec.md

MalloZup:
there is a broken link on design page

stefanotorresi:
fixed!

pacemaker_metrics.go
```diff
@@ -61,11 +66,136 @@ type resource struct {
...
+		"nodes":           NewMetricDesc("pacemaker", "nodes", "The nodes in the cluster; one line per name, per status", []string{"name", "type", "status"}),
+		"nodes_total":     NewMetricDesc("pacemaker", "nodes_total", "Total number of nodes in the cluster", nil),
+		"resources":       NewMetricDesc("pacemaker", "resources", "The resources in the cluster; one line per id, per status", []string{"node", "id", "role", "managed", "status"}),
+		"resources_total": NewMetricDesc("pacemaker", "resources_total", "Total number of resources in the cluster", nil),
+	}
+
+	crmMonPath = "/usr/sbin/crm_mon"
+)
+
+func NewPacemakerCollector() (*pacemakerCollector, error) {
```
MalloZup: regarding the Uppercase exported public functions, in theory since we are using only one package they aren't needed to be exported. Imho we should perhaps don't `Foo` but `foo` since we don't plan to expose them as public API really but I'm ok also like this.

drbd_metrics_test.go
```diff
@@ -5,104 +5,111 @@ import (
...
-				  "pending": 0,
-				  "unacked": 0,
-				  "has-sync-details": false,
-				  "has-online-verify-details": false,
-				  "percent-in-sync": 100.00
-				} ]
-			} ]
-		}]`)
+	var drbdDataRaw = []byte(`
+[
```
MalloZup: why this doesn't go to fixtures?

pacemaker_metrics.go
stefanotorresi: actually I guess in the future we'll likely move the various subsystems into separate packages, because we already have some name clashes with the XML parsing structs

drbd_metrics_test.go
stefanotorresi: I didn't really change this particular test, I only indented the JSON, but I guess we could move this to a `.json` file and use the same for both this test and the one that invokes the fake executable. Do you want me to do that?

MalloZup: yep. I think it is more consistent with other fixture metrics imho as approach. We can do also as follow-up or you can do it while I review :grin:

pacemaker_metrics.go
MalloZup: Ok let see later. was a historic remark for moment :grin:

doc/metric_spec.md
```diff
@@ -1,112 +1,179 @@
...
-they can be either set to `1` or they are absent, this is because they track the real state of the cluster resources monitored.
+## Pacemaker 
 
-1. [ha_cluster_node_resources](#ha_cluster_node_resources)
-2. [ha_cluster_nodes](#ha_cluster_nodes)
-3. [ha_cluster_nodes_configured_total](#ha_cluster_nodes_configured_total)
-4. [ha_cluster_resources_configured_total](#ha_cluster_resources_configured_total)
+The Pacemaker subsystem collects an atomic snapshot of the HA cluster directly from the XML CIB of Pacemaker via `crm_mon`.
 
+0. [Sample](../test/pacemaker.metrics)
```
MalloZup: perhaps here:

instead of put the 0. `sample`, I would perhaps move it at the end of the index with  a link `examples`

1. foo
2. bar
3. baf

examples: (link)

MalloZup:
@stefanotorresi  can you remove some commits?

Like the `revert` thing we don't really need it same as adding stonith metric also revert things.

It would benficial for us to have less commits which are more relevant to the actual feature. I'm not that obsessed but we have to much commits which could be removed in this PR.

stefanotorresi:
@MalloZup if I rewrite the commit history again I'm gonna have to rebase the other two PRs all over again, are you sure it's needed? Changes are not trivial and there would be a lot of conflicts to resolve.

PR #67, opened 2019-10-21
Feature: metric timestamps

stefanotorresi:
follow up from #66:

add the `timestamp` field to all the metrics, as specified by the [metrics line protocol](https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md#text-format-details)


diegoakechi:
Adding @ne-msft here so he can get aware of this change.

PR #75, opened 2019-10-25
Add delta metric: Percent in sync

MalloZup:
# Info:

actually here is only the last commit relavant.


stefanotorresi:
I think this should be implemented in a separated metric, not in a label.

E.g.
`ha_cluster_drbd_connections_sync{peer_node_id=foo resource=bar} 100 1234`

Otherwise, we would actually be creating _hundreds_ of new time series for each connection, while we only want one per connection (as many as connections).

MalloZup:
Make sense yep

PR #80, opened 2019-10-28
implement resource_agent_changes metric

MalloZup:
# desc

Implement metric for track changes on cluster resource agent

An user will need to use some promql function like `delta` or `changes` for using this metric

### DIFF
--- a/doc/metric_spec.md
+++ b/doc/metric_spec.md
@@ -31,7 +31,7 @@ The Pacemaker subsystem collects an atomic snapshot of the HA cluster directly f
 5. [`ha_cluster_pacemaker_stonith_enabled`](#ha_cluster_pacemaker_stonith_enabled)
 6. [`ha_cluster_pacemaker_fail_count`](#ha_cluster_pacemaker_fail_count)
 7. [`ha_cluster_pacemaker_migration_threshold`](#ha_cluster_pacemaker_migration_threshold)
-8. [`ha_cluster_pacemaker_resource_configuration_changes`](#ha_cluster_pacemaker_resource_configuration_changes)
+8. [`ha_cluster_pacemaker_config_last_change`](#ha_cluster_pacemaker_config_last_change)
 
 ### `ha_cluster_pacemaker_nodes`
 
@@ -102,12 +102,12 @@ The value can vary from 0, 1 , 5 etc to `+Inf`, that  correspond to the infinity
 The number of migration threshold pro node and resource ID set by a pacemaker cluster. 
 Possible values are positive numbers.
 
-### `ha_cluster_pacemaker_resource_configuration_changes`
+### `ha_cluster_pacemaker_config_last_change`
 
 #### Description
 
-This metric return a unix timestamp which correspond to the C.ansi timestamp updated by pacemaker in cluster.
-It is useful to see if the resource agent configuration in the cluster has changed over a period of time and gets alerts.
+The relevant part of this metric is its timestamp, which corresponds to the last time Pacemaker configuration changed.
+The actual metric value will always be 1 and can be ignored.
 
 ## Corosync
 
--- a/doc/metric_spec.md
+++ b/doc/metric_spec.md
@@ -31,7 +31,7 @@ The Pacemaker subsystem collects an atomic snapshot of the HA cluster directly f
 5. [`ha_cluster_pacemaker_stonith_enabled`](#ha_cluster_pacemaker_stonith_enabled)
 6. [`ha_cluster_pacemaker_fail_count`](#ha_cluster_pacemaker_fail_count)
 7. [`ha_cluster_pacemaker_migration_threshold`](#ha_cluster_pacemaker_migration_threshold)
-
+8. [`ha_cluster_pacemaker_resource_configuration_changes`](#ha_cluster_pacemaker_resource_configuration_changes)
 
 ### `ha_cluster_pacemaker_nodes`
 
@@ -102,6 +102,12 @@ The value can vary from 0, 1 , 5 etc to `+Inf`, that  correspond to the infinity
 The number of migration threshold pro node and resource ID set by a pacemaker cluster. 
 Possible values are positive numbers.
 
+### `ha_cluster_pacemaker_resource_configuration_changes`
+
+#### Description
+
+This metric return a unix timestamp which correspond to the C.ansi timestamp updated by pacemaker in cluster.
+It is useful to see if the resource agent configuration in the cluster has changed over a period of time and gets alerts.
 
 ## Corosync
 
--- a/pacemaker_metrics.go
+++ b/pacemaker_metrics.go
@@ -82,13 +82,14 @@ var (
 	pacemakerMetrics = metricDescriptors{
 		// the map key will function as an identifier of the metric throughout the rest of the code;
 		// it is arbitrary, but by convention we use the actual metric name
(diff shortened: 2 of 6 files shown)


test/fake_crm_mon.sh
```diff
@@ -84,7 +84,8 @@ cat <<EOF
     </node_history>
     <tickets>
     </tickets>
-    <bans>
+      <bans>
+        <ban id="cli-ban-msl_SAPHana_PRD_HDB00-on-damadog-hana01" resource="msl_SAPHana_PRD_HDB00" node="damadog-hana01" weight="-1000000" master_only="false" />
```
MalloZup: this is unreleated to this pr but it will be usefull for next metric. we can keep it . is not arm

doc/metric_spec.md
```diff
@@ -102,6 +102,12 @@ The value can vary from 0, 1 , 5 etc to `+Inf`, that  correspond to the infinity
 The number of migration threshold pro node and resource ID set by a pacemaker cluster. 
 Possible values are positive numbers.
 
+### `ha_cluster_pacemaker_resource_configuration_changes`
```
stefanotorresi: I would suggest changing the name from `resource_configuration_changes` to `last_configuration_change` or `last_cib_change`, as this is not really resource-related, but it's cluster-wide.

pacemaker_metrics.go
```diff
@@ -237,6 +243,16 @@ func (c *pacemakerCollector) recordFailCountMetrics(pacemakerStatus pacemakerSta
...
 }
 
+func (c *pacemakerCollector) recordResourceAgentsChanges(pacemakerStatus pacemakerStatus, ch chan<- prometheus.Metric) {
+	t, err := time.Parse(time.ANSIC, pacemakerStatus.Summary.LastChange.Time)
+	if err != nil {
+		log.Warnln(err)
+		return
+	}
+	// is the resource have changed we set a different timeout from pacemaker
+	ch <- c.makeGaugeMetric("resource_configuration_changes", float64(t.Unix()))
```
stefanotorresi: I'm not super convinced on how we are modelling this metric, to be frank.
IMHO, representing the time in a float, when we already have an integer timestamp at the end of the line, is super-confusing for the user. One possible solution would be to always set the value to `0`, and use the timestamp itself.

eg:
```suggestion
	ch <- prometheus.NewMetricWithTimestamp(t, prometheus.MustNewConstMetric(c.metrics["resource_configuration_changes"], prometheus.CountValue, 0))
```

What are your thoughts on this?

test/fake_crm_mon.sh
stefanotorresi: no problemo

pacemaker_metrics.go
MalloZup: I agree that is a tricky metrics :grin: . let mes seartch about the timestamp.

Our goal is to indicate a change so an user can detect that with promql functions.

doc/metric_spec.md
gao-yan: I agree. Something like `ha_cluster_pacemaker_last_config_change`.

doc/metric_spec.md
```diff
@@ -102,6 +102,12 @@ The value can vary from 0, 1 , 5 etc to `+Inf`, that  correspond to the infinity
...
+
+#### Description
+
+This metric return a unix timestamp which correspond to the C.ansi timestamp updated by pacemaker in cluster.
+It is useful to see if the resource agent configuration in the cluster has changed over a period of time and gets alerts.
```
gao-yan: We just call it "resource configuration" rather than "resource agent configuration". A resource is of course managed via an agent.

doc/metric_spec.md
MalloZup: done thx

gao-yan: Doesn't real matter, or `ha_cluster_pacemaker_config_last_change` to be object-oriented :-)

pacemaker_metrics.go
```diff
@@ -237,6 +243,16 @@ func (c *pacemakerCollector) recordFailCountMetrics(pacemakerStatus pacemakerSta
...
 }
 
+func (c *pacemakerCollector) recordResourceAgentsChanges(pacemakerStatus pacemakerStatus, ch chan<- prometheus.Metric) {
+	t, err := time.Parse(time.ANSIC, pacemakerStatus.Summary.LastChange.Time)
+	if err != nil {
+		log.Warnln(err)
+		return
+	}
+	// is the resource have changed we set a different timeout from pacemaker
+	ch <- prometheus.NewMetricWithTimestamp(t, prometheus.MustNewConstMetric(c.metrics["ha_cluster_pacemaker_config_last_change"], prometheus.CounterValue, 1))
```
MalloZup: instead of 0 , is better to use `1` since it will be more inline with our exporter design with 1 as special value of being present

stefanotorresi: good point!

stefanotorresi: mind that the `c.metrics` lookups should not include the namespace and the subsystem, so in this case it would be `c.metrics["config_last_change"]`

MalloZup: good yeah typ;o

go.mod
```diff
@@ -3,6 +3,8 @@ module github.com/ClusterLabs/ha_cluster_exporter
 go 1.12
 
 require (
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.1 // indirect
```
stefanotorresi: probably need a run of `go mod tidy` here

doc/metric_spec.md
```diff
@@ -102,6 +102,13 @@ The value can vary from 0, 1 , 5 etc to `+Inf`, that  correspond to the infinity
 The number of migration threshold pro node and resource ID set by a pacemaker cluster. 
 Possible values are positive numbers.
 
+### `ha_cluster_pacemaker_config_last_change`
+
+#### Description
+
+This metric has 2 values: the first one is always set to 1 which indicate the presence of the metric. 
+The second one, is the timestamp that correspond to the timestamp updated by pacemaker forresource configuration last change.
+It is useful to see if the resource configuration in the cluster has changed over a period of time and gets alerts.
```
stefanotorresi: I propose rewording this because every metric has the timestamp, and this is no exception. Also, this metric will always be present.

Maybe something like:
```
The relevant part of this metric is its timestamp, which corresponds to the last time Pacemaker configuration changed.
The actual metric value will always be 1 and can be ignored.
```

pacemaker_metrics.go
```diff
@@ -237,6 +243,16 @@ func (c *pacemakerCollector) recordFailCountMetrics(pacemakerStatus pacemakerSta
 	}
 }
 
+func (c *pacemakerCollector) recordResourceAgentsChanges(pacemakerStatus pacemakerStatus, ch chan<- prometheus.Metric) {
+	t, err := time.Parse(time.ANSIC, pacemakerStatus.Summary.LastChange.Time)
+	if err != nil {
+		log.Warnln(err)
+		return
+	}
+	// is the resource have changed we set a different timeout from pacemaker
```
stefanotorresi: This comment is a bit unclear :stuck_out_tongue_winking_eye: 

```suggestion
	// The time of the last change is used to set the metric timestamp. The value is always 1 and is not relevant.
```

doc/metric_spec.md
MalloZup: done thx

MalloZup:
@stefanotorresi great thx for review.

Issue #81, opened 2019-10-29
Feature Request: Implement listen IPv4 option 


ashleyprimo:
Currently, it is not possible to alter which address the exporter will listen on; I propose the addition of a `-ip4` flag.

E.g. `./ha_cluster_exporter --ip4 127.0.0.1`

Pull request to follow :)

Issue #27, opened 2019-10-11
Feature Request: Implement Log Levelling 


ashleyprimo:
It would be very nice to see the implementation of _log levelling_ to enable specification of log levels that are output which is a useful feature to have in use within a production environment. 

In the current operation, the exporter continually output's *info* level logs.

**Example**:
```
2019/10/11 11:10:10 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:10:10 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:15 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:15 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:10:20 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:20 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:10:25 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:25 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:10:30 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:30 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:10:35 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:35 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:10:40 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:40 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:10:45 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:45 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:10:50 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:50 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:10:55 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:10:55 [INFO]: Reading quorum status with corosync-quorumtool...
2019/10/11 11:11:00 [INFO]: Reading cluster configuration with crm_mon..
2019/10/11 11:11:00 [INFO]: Reading quorum status with corosync-quorumtool...
[...]

PR #82, opened 2019-10-29
Add -address listen, log -level flag

ashleyprimo:
Per issue #81 - add the ability to listen on a specific IPv4 address.

### DIFF
--- a/ha_cluster_exporter.go
+++ b/ha_cluster_exporter.go
@@ -73,14 +73,33 @@ func landingpage(w http.ResponseWriter, r *http.Request) {
 		</html>`))
 }
 
+func loglevel(opt string) {
+	switch opt {
+	case "error":
+		log.SetLevel(log.ErrorLevel)
+	case "warn":
+		log.SetLevel(log.WarnLevel)
+	case "info":
+		log.SetLevel(log.InfoLevel)
+	case "debug":
+		log.SetLevel(log.DebugLevel)
+	default:
+		log.Warnln("Unrecognized log level, default to `info` level")
+	}
+}
+
 var (
 	clock      Clock = &SystemClock{}
 	portNumber       = flag.String("port", "9002", "The port number to listen on for HTTP requests.")
+	address          = flag.String("address", "0.0.0.0", "The address to listen on for HTTP requests.")
+	logLevel         = flag.String("level", "info", "The level of logs to log")
 )
 
 func main() {
 	// read cli option and setup initial stat
 	flag.Parse()
+
+	loglevel(*logLevel)
 
 	pacemakerCollector, err := NewPacemakerCollector()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ashleyprimo:
Second commit also add feature per the last comment on #27

ha_cluster_exporter.go
```diff
@@ -73,15 +73,33 @@ func landingpage(w http.ResponseWriter, r *http.Request) {
...
+		log.SetLevel(log.InfoLevel)
+	case "debug":
+		log.SetLevel(log.DebugLevel)
+	}
+}
+
 var (
 	clock      Clock = &SystemClock{}
 	portNumber       = flag.String("port", "9002", "The port number to listen on for HTTP requests.")
+	ip4              = flag.String("ip4", "0.0.0.0", "The IP to listen on for HTTP requests.")
```
MalloZup: I'm thinking here, why should we call it `ip4`? 

To me it sounds perhaps to specific. Looking at the code, we could call it `ip`.

Because if we call it `ip4` an user would expect then to have a `ip6` flag, which would just duplicate stuff.

ha_cluster_exporter.go
```diff
@@ -73,15 +73,33 @@ func landingpage(w http.ResponseWriter, r *http.Request) {
...
+	logLevel         = flag.String("level", "info", "The level of logs to log")
 )
 
 func main() {
 	// read cli option and setup initial stat
 	flag.Parse()
 
+	// Log Level
```
MalloZup: this comment is redundant, I can read it from the function that we are setting the logLevel :grin: 

either you remove it, or you add more context.

Like 

```setup loglevel accordingly from user CLI flag , default is info. ```

But again, Perhaps we should just remove it.

File: ha_cluster_exporter.go [hunk 3]
```diff
@@ -73,15 +73,33 @@ func landingpage(w http.ResponseWriter, r *http.Request) {
...
+	switch opt {
+	case "error":
+		log.SetLevel(log.ErrorLevel)
+	case "warn":
+		log.SetLevel(log.WarnLevel)
+	case "info":
+		log.SetLevel(log.InfoLevel)
+	case "debug":
+		log.SetLevel(log.DebugLevel)
+	}
```
MalloZup: we miss here a `default` clause.
something like 
``
default:
		return errors.Errorf("unrecognized log level %q", s)
}
```

File: ha_cluster_exporter.go (hunk 1 above)
Comment (stefanotorresi) [reply ha_cluster_exporter.go]: I would add that we shouldn't distinguish between ip4 and ip6 because `net.Dial()` (which is the socket backend eventually used by `http.ListenAndServe`) [supports both](https://golang.org/pkg/net/#Dial).
Also, I suggest calling this just `address`, because it doesn't necessarily need to be an IP, it could also be a hostname.

File: ha_cluster_exporter.go (hunk 2 above)
Comment (ashleyprimo) [reply ha_cluster_exporter.go]: Yes, I did intend to add some more context- forgot to do this :laughing:

File: ha_cluster_exporter.go (hunk 1 above)
Comment (ashleyprimo) [reply ha_cluster_exporter.go]: Yes, that's a really good point, I'll change this.

File: ha_cluster_exporter.go (hunk 3 above)
Comment (ashleyprimo) [reply ha_cluster_exporter.go]: Thinking of something such as:
```
        default:
                log.SetLevel(log.InfoLevel)
                log.Warnln("Unrecognized log level, default to `info` level")
        }
```

ashleyprimo: This said we can do away with `log.SetLevel(log.InfoLevel)` since default behaviour is to use `info` level

PR #83, opened 2019-10-29
Implement ban constraint metric

MalloZup:
implement a metric for ban constraints.

```

# HELP ha_cluster_pacemaker_constraints Indicate if some constraints of different type is present on cluster for resource ID
# TYPE ha_cluster_pacemaker_constraints gauge
ha_cluster_pacemaker_constraints{id="cli-ban-msl_SAPHana_PRD_HDB00-on-damadog-hana01",type="ban"} 1 1234
```

### DIFF
--- a/pacemaker_metrics.go
+++ b/pacemaker_metrics.go
@@ -29,6 +29,11 @@ type pacemakerStatus struct {
 			} `xml:"resource_history"`
 		} `xml:"node"`
 	} `xml:"node_history"`
+	Bans struct {
+		Ban []struct {
+			ID string `xml:"id,attr"`
+		} `xml:"ban"`
+	} `xml:"bans"`
 }
 
 type summary struct {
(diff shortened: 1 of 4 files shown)


pacemaker_metrics.go
```diff
@@ -93,7 +98,8 @@ var (
 		"stonith_enabled":     NewMetricDesc("pacemaker", "stonith_enabled", "Whether or not stonith is enabled", nil),
 		"fail_count":          NewMetricDesc("pacemaker", "fail_count", "The Fail count number per node and resource id", []string{"node", "resource"}),
 		"migration_threshold": NewMetricDesc("pacemaker", "migration_threshold", "The migration_threshold number per node and resource id", []string{"node", "resource"}),
-		"config_last_change":  NewMetricDesc("pacemaker", "config_last_change", "Indicate if a configuration of resource has changed in cluster", []string{}),
+		"config_last_change":  NewMetricDesc("pacemaker", "config_last_change", "Indicate if a configuration of resource has changed in cluster", nil),
+		"constraints":         NewMetricDesc("pacemaker", "constraints", "Indicate if a constraints of specific type is present per resource ID", []string{"type", "id"}),
```
stefanotorresi: I think the help here is a little bit misleading: the `id` label is actually the id of the constraint, not of the resource.
I know that the ban id usually also contains the resource id, but after all, it's an opaque value so I think we should add the `resource` label anyway, and distinguish the two so that users can join `contraints` and `resources` metrics.

MalloZup:
@stefanotorresi  I have adressed your comment on the new PR. Initially I wanted to do 2 prs but I have managed to do 1 pr for both metrics types. (PREFER/BAN). Closing this since the new one contains latest changes.

Issue #87, opened 2019-10-31
The cluster binaries locations varies depending on Releases or Distributions


diegoakechi:
Depending on distro or on the service pack, the cluster binaries locations can vary, causing the exporter to not find the executable. For example, between SLE-12-SP3 and SLE-15, the binary `drbdsetup` has changed location and the exporter shows the following error:

```Oct 31 13:32:14 dev-1 ha_cluster_exporter[73317]: time="2019-10-31T13:32:14Z" level=warning msg="Could not initialize DRBD collector: '/usr/sbin/drbdsetup' not found: stat /usr/sbin/drbdsetup: n...r directory\n"```

And here we see the real location 
```
dev-1:~ # which drbdsetup
/sbin/drbdsetup
```

The exporter should allow different locations and ideally should try to find it by itself.

**Comment by ashleyprimo** [comment]
This could be solved using something such as `path, err := exec.LookPath("binary")`- I'll submit a PR for this (likely on Monday) if this is still open 👍

**Comment by diegoakechi** [comment]
Great @ashleyprimo. Thanks!

**Comment by MalloZup** [comment]
regarding PR and issue, I'm not sure if we are embarking on the good direction. Imho we should just use `command.exec`. https://golang.org/pkg/os/exec/#Command

In fact we are doing what already the command do , if you look at upstream doc: 

```
If name contains no path separators, Command uses LookPath to resolve name to a complete path if possible. Otherwise it uses name directly as Path. ```

IMHO the best approach would to remove the OS.stat and check if binary, steps and execute the cmd.exec, which will already uses LookPat.  If this fail, we can just adapt the error message that either binary was not found or the binary whatsoever wasn't executable. But this will allow us to be more clean without to many checks which we can spare imho in this scenario. (since cmd.exec uses already)

ashleyprimo:
This is a very good point, makes sense that it would revert to `LookPath` natively; however, if were to just rely on the `exec` we would lose the ability to check if it's able to find the binary and if the binary is executable - since `os.Stat` relies on an absolute path.

Unless I'm missing something? :smiley:

MalloZup:
ok but on highlevel point this 2 check what they would provide? if we can't execute the binary (because of lack of findings or is not exec) we have an IF which check the error , and  an error is throwed

PR #88, opened 2019-11-03
Implement binary lookup

ashleyprimo:
This is a _fix_ / improvement per issue #87 - fix itself is working, however, discovered that my changes conflict with testing.

So this is not yet ready to merge until this is solved, and testing is fixed - I welcome any input on the best course for testing and moreover, my suggested changed  :slightly_smiling_face:

### DIFF
--- a/pacemaker_metrics.go
+++ b/pacemaker_metrics.go
@@ -101,19 +101,19 @@ var (
 		"constraints":         NewMetricDesc("pacemaker", "constraints", "Indicate if a constraints of specific type is present per ID and per resource", []string{"type", "id", "resource"}),
 	}
 
-	pacemakerTools = map[string]string{
+	pacemakerBinaries = map[string]string{
 		"crmMon":   "crm_mon",
 		"cibAdmin": "cibadmin",
 	}
 )
 
 func NewPacemakerCollector() (*pacemakerCollector, error) {
-	for tool, toolBinary := range pacemakerTools {
+	for tool, toolBinary := range pacemakerBinaries {
 		binaryPath, err := exec.LookPath(toolBinary)
 		if err != nil {
 			return nil, errors.Wrapf(err, "'%s' not found", toolBinary)
 		} else {
-			pacemakerTools[tool] = binaryPath
+			pacemakerBinaries[tool] = binaryPath
 		}
 
 		fileInfo, err := os.Stat(binaryPath)
@@ -163,7 +163,7 @@ func (c *pacemakerCollector) Collect(ch chan<- prometheus.Metric) {
 
 func getPacemakerStatus() (pacemakerStatus, error) {
 	var pacemakerStatus pacemakerStatus
-	pacemakerStatusXML, err := exec.Command(pacemakerTools["crmMon"], "-X", "--group-by-node", "--inactive").Output()
+	pacemakerStatusXML, err := exec.Command(pacemakerBinaries["crmMon"], "-X", "--group-by-node", "--inactive").Output()
 	if err != nil {
 		return pacemakerStatus, errors.Wrap(err, "error while executing crm_mon")
 	}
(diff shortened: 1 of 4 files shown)


sbd_metrics.go
```diff
@@ -24,20 +24,28 @@ var (
 	}
 
 	sbdConfigPath = "/etc/sysconfig/sbd"
-	sbdPath       = "/usr/sbin/sbd"
+	sbdTools      = map[string]string{
```
MalloZup: I would not add `tools` everywhere

pacemaker_metrics.go
```diff
@@ -101,21 +101,24 @@ var (
 		"constraints":         NewMetricDesc("pacemaker", "constraints", "Indicate if a constraints of specific type is present per ID and per resource", []string{"type", "id", "resource"}),
 	}
 
-	crmMonPath   = "/usr/sbin/crm_mon"
-	cibAdminPath = "/usr/sbin/cibadmin"
+	pacemakerTools = map[string]string{
```
MalloZup: I would rename it with pacemakerBinaries

sbd_metrics.go
ashleyprimo: > thx @ashleyprimo . the `exec.LookPath` is the way to go. 
> 
> Can you describe what is the problem with the tests , without that I have to debug? I just have looked at the pr on highlevel pov

So the issue in respect of testing; is that we define tests like so:
```        corosyncTools["cfgtool"] = "test/dummy"
        corosyncTools["quorumtool"] = "test/fake_corosync-quorumtool.sh"

        _, err := NewCorosyncCollector()
```

So the issue is that now if you fed `corosyncTools["cfgtool"] = "test/dummy"` you will not be able to derive the error:
```        if err.Error() != "'test/dummy' is not executable" {
                t.Errorf("Unexpected error: %v", err)
        }
```

Since `test/dummy` script will never be found; the same can be said for `test/fake_corosync-quorumtool.sh` - this code is how much of testing is done. 

This said just thinking about this; if we could just set the $PATH environment variable to `/test` in the CI, and `corosyncTools["cfgtool"] = "test/dummy"` to `corosyncTools["cfgtool"] = "dummy"` this could work; however not ever attempted this with TravisCI - so no idea if this is possible...

ashleyprimo:
> thx @ashleyprimo . the `exec.LookPath` is the way to go. 
> 
> Can you describe what is the problem with the tests , without that I have to debug? I just have looked at the pr on highlevel pov

So the issue in respect of testing; is that we define tests like so:
```        corosyncTools["cfgtool"] = "test/dummy"
        corosyncTools["quorumtool"] = "test/fake_corosync-quorumtool.sh"

        _, err := NewCorosyncCollector()
```

So the issue is that now if you fed `corosyncTools["cfgtool"] = "test/dummy"` you will not be able to derive the error:
```        if err.Error() != "'test/dummy' is not executable" {
                t.Errorf("Unexpected error: %v", err)
        }
```

Since `test/dummy` script will never be found; the same can be said for `test/fake_corosync-quorumtool.sh` - this code is how much of testing is done. 

This said just thinking about this; if we could just set the $PATH environment variable to `/test` in the CI, and `corosyncTools["cfgtool"] = "test/dummy"` to `corosyncTools["cfgtool"] = "dummy"` this could work; however not ever attempted this with TravisCI - so no idea if this is possible...

sbd_metrics.go
ashleyprimo: I imagine you mean, to change the wording- if so done :smiley:

MalloZup:
@ashleyprimo  thx for PR. I have update the issue. IMho this approach might be evaluated. To me we are adding to many steps when we could avoid them and let just simply fail it without this . see my comments on issue

stefanotorresi:
I have given some thought to this, but I haven't yet come up with a definitive solution.
The first thing I thought to fix #87, is using a configuration file using a common format like `.ini`, `.yaml` or something like that, and set the various path there (with common safe defaults).
I also thought about using Go ldflags during the build, and ship the binary with the correct path pre-baked in.
I must say I don't particularly like using `PATH` lookup, because the environment is not something we can always control easily, and it introduces documentation overhead; i.e. we should be able to just clone the project and run `go test` without changing the environment.

ashleyprimo:
Just a thought, we could try something such as the following:
   1. Check the proposed binary against `/test` using `os.stat` (this can fail silently)
   2. If file not found; default to `exec` without a path (invoking lookup)

It's kind of 'hacky' however it would solve both issues, testing, in the same way, would be obtainable as would the location of the binary.  

This said, a configuration would be a good idea- but in the aim of minimising the complexity of the code.

In regards to `PATH` lookup; really if a program is installed correctly- then it should be able to be found in one of the `PATH` directories; if it can't then it's likely that the program in question was not installed correctly which does not constitute an issue with the exporter :slightly_smiling_face:

stefanotorresi:
hey @ashleyprimo look at #91 and let me know if you find it satisfactory!

MalloZup:
closing since we fixed on the new pr . @ashleyprimo thx a lot for your time and help on this. Wish you a great day

PR #91, opened 2019-11-07
RFC: add configuration via files and additional CLI flags

### LINKED ISSUE #87: The cluster binaries locations varies depending on Releases or Distributions

stefanotorresi:
This PR illustrates my proposal to allow changing some parameters that were so far hard-coded in the code: namely, the various paths of the external executables the project depends on (e.g. `crm_mon`, `corosyng-cfgtool`, etc.). This is mainly to fix #87.

## The solution: multiple configuration mechanisms

The exporter can now be configured via CLI flags and/or via static configuration files. This has been implemented with the popular library [spf13/viper](//github.com/spf13/viper), with minimal effort. In the future, we could also enable the support for environment variables, quite handy if we were to ship a Docker image.

An example YAML configuration is now provided in the repo:
https://github.com/ClusterLabs/ha_cluster_exporter/blob/48c40c52331b0d3de7e09063863ac33c18ba36a0/ha_cluster_exporter.yaml#L1-L11

This shall be used as a starting point when building the various RPM packages with OBS, to ship distro-specific configs inside `/usr/etc`, [which has been recently designated as the conventional location for default configurations provided by SUSE and RedHat package maintainers](https://github.com/thkukuk/atomic-updates_and_etc).

In addition to this mechanism, all these parameters are now exposed via flags:
```
Usage of ./ha_cluster_exporter:
      --address string                     The address to listen on for HTTP requests (default "0.0.0.0")
      --cibadmin-path string               path to cibadmin executable (default "/usr/sbin/cibadmin")
      --corosync-cfgtoolpath-path string   path to corosync-cfgtool executable (default "/usr/sbin/corosync-cfgtool")
      --corosync-quorumtool-path string    path to corosync-quorumtool executable (default "/usr/sbin/corosync-quorumtool")
      --crm-mon-path string                path to crm_mon executable (default "/usr/sbin/crm_mon")
      --drbdsetup-path string              path to drbdsetup executable (default "/usr/sbin/drbdsetup")
      --log-level string                   The minimum logging level; levels are, in ascending order: debug, info, warn, error (default "info")
      --port string                        The port number to listen on for HTTP requests (default "9002")
      --sbd-config-path string             path to sbd configuration (default "/etc/sysconfig/sbd")
      --sbd-path string                    path to sbd executable (default "/usr/sbin/sbd")
```

These could be leveraged by the systemd unit file, if users so desire.

## Implementation notes

The most prominent changes required in the code involve the fact that the paths are now locally injected inside the collectors, instead of being stored in global variables; this was actually a conscious design oversight that was always intended to be addressed in some way, sooner or later.

The actual integration with CLI flags and file configuration only takes a few declarations inside the root `init()` function:
https://github.com/ClusterLabs/ha_cluster_exporter/blob/48c40c52331b0d3de7e09063863ac33c18ba36a0/ha_cluster_exporter.go#L112-L134

The default behaviour remains unchanged, but users can now optionally store a configuration file, in various supported formats, inside either the current working directory (`./`), the user's `$HOME/.config/` directory, or the system's `/etc/` and `/usr/etc` directories, (all in this exact order, i.e. the first match has the priority).

While I was at it, I also took the chance to move the often duplicated executable validation logic into a shared function, invoked by the various collector constructors:
https://github.com/ClusterLabs/ha_cluster_exporter/blob/48c40c52331b0d3de7e09063863ac33c18ba36a0/ha_cluster_exporter.go#L63-L77

## Breaking changes

These changes actually break all the previous flags, which were registered via a single dash (e.g. `-port` is now `--port`). This is required by the [spf13/pflag](//github.com/spf13/pflag) wrapper, superseding Go stdlib's `flags`, which also opens the possibility to use shorthand flags.

## Other possible solutions

The other solutions evaluated to solve the same problem where the following:
- **Dynamic lookup of the files via the `PATH` environment variable.**  
  This would have been a very simple implementation, but somewhat limited in flexibility, and it didn't support the current testing setup, which has some fake tools in an arbitrary directory that doesn't live in the PATH. It would also not support non-executable configuration files that we need to inspect like `/etc/sysconfig/sbd`.
- **Directly build the parameters via Go ldflags.**  
  While technically possible, this is actually a very bad idea: it would introduce complexity in the build pipeline and produce binaries that could not be simply ported from one environment to another, defeating one of the Go strongest points, i.e. being able to ship a single, static, portable binary.


diegoakechi:
I like the approach and it looks clean to me. @ashleyprimo You may want to check this and give your comments too...

corosync_metrics.go
```diff
@@ -21,33 +20,27 @@ var (
...
-		if os.IsNotExist(err) {
-			return nil, errors.Wrapf(err, "'%s' not found", toolPath)
-		}
-		if (fileInfo.Mode() & 0111) == 0 {
-			return nil, errors.Errorf("'%s' is not executable", toolPath)
-		}
+func NewCorosyncCollector(cfgToolPath string, quorumToolPath string) (*corosyncCollector, error) {
+	err := CheckExecutables(cfgToolPath, quorumToolPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "external executable check failed")
```
MalloZup: are we errorring here much generic? 

If we are wrapping error we should add  a more precise context on corosync

ha_cluster_exporter.go
```diff
@@ -57,6 +59,20 @@ func NewMetricDesc(subsystem, name, help string, variableLabels []string) *prome
 	return prometheus.NewDesc(prometheus.BuildFQName(NAMESPACE, subsystem, name), help, variableLabels, nil)
 }
 
+func CheckExecutables(paths ...string) error {
+	// check that all executables we rely on exist and are executables
```
MalloZup: `nit` inn a `golint` way of thing the comment should be like
functioname and desc
\\ CheckExecutables  check that all executables we rely on exist and are executables
func foof

stefanotorresi: I don't quite understand what you mean. Do you want the comment moved above the declaration?

MalloZup:
@stefanotorresi  great job. I really like this. :clap: :clap: 

At first glance seems clean. I will take a deeper look right now

I was thinking if we are perhaps missing some documentation about the fact that exporter is configurable.

And also perhaps we might need also to adapt the systemd services file. (to me this could be either follow-up issue or done directly)

stefanotorresi:
> I was thinking if we are perhaps missing some documentation about the fact that exporter is configurable.

Yeah, you're quite right! I'll will add something to the README.

> And also perhaps we might need also to adapt the systemd services file. (to me this could be either follow-up issue or done directly)

Not really, I think the unit file is okay, unless I'm missing something. We instead need to change the obs `.spec` to create the dedicated configs for older versions of SLES, but I would do that in a different PR.

corosync_metrics.go
stefanotorresi: The message of the wrapped error will also show which executable is failing, so it will not be that generic! This is an example of the final result when launching the exporter:

```
$ ./ha_cluster_exporter 
INFO[0000] Using config file: /home/etienne/work/SUSE/ha_cluster_exporter/ha_cluster_exporter.yaml 
WARN[0000] Could not initialize Pacemaker collector: external executable check failed: stat /usr/sbin/crm_mon: no such file or directory 
WARN[0000] Could not initialize Corosync collector: external executable check failed: stat /usr/sbin/corosync-cfgtool: no such file or directory 
WARN[0000] Could not initialize SBD collector: external executable check failed: stat /usr/sbin/sbd: no such file or directory 
WARN[0000] Could not initialize DRBD collector: external executable check failed: stat /usr/sbin/drbdsetup: no such file or directory 
INFO[0000] Serving metrics on 0.0.0.0:9002 
```

I think it's verbose enough, no?

ha_cluster_exporter.go
MalloZup: @stefanotorresi take this as example:
https://github.com/kubernetes/kubernetes/blob/master/cmd/genutils/genutils.go#L25

MalloZup: for a public function in go normally the convention is:

`// funcname blalbla`
func foo

ashleyprimo:
Yep- this is very nice; solves the issue very nicely :smiley:

ha_cluster_exporter.go
```diff
@@ -84,53 +100,92 @@ func loglevel(opt string) {
...
+
+	err := config.BindPFlags(flag.CommandLine)
+	if err != nil {
+		log.Errorf("Could not bind config to CLI flags: %v", err)
+	}
+}
 
 func main() {
-	// read cli option and setup initial stat
+	var err error
```
MalloZup: why this duplicata? we can delete this line and just add on line 137 `:=`

corosync_metrics.go
MalloZup: yeah i resolved. ignore this.

ha_cluster_exporter.go
stefanotorresi: this was actually done on purpose: since the main has gotten very long, declaring `err` at the top avoids errors in case the first assignment is moved around.

stefanotorresi:
@MalloZup @diegoakechi I've updated [the readme](https://github.com/ClusterLabs/ha_cluster_exporter/blob/e3650254c9c60f291e545195f42e4fdc1d6ff87b/README.md), added support for the `/usr/etc`, and updated the text of the RFC in the opening post.

Ready for another round of reviews.

README.md
```diff
@@ -2,48 +2,107 @@
...
+You can deploy a full HA Cluster via Terraform with [SUSE/ha-sap-terraform-deployments](https://github.com/SUSE/ha-sap-terraform-deployments).
+
+### Configuration
+
+All the runtime parameters can be configured either via CLI flags or via a configuration file, both or which are completely optional.
+
+For more details, refer to the help message via `ha_cluster_exporter --help`.
+
+**Note**:
+the built-in defaults are tailored for the latest version of SUSE Linux Enterprise.
```
MalloZup: should we also add `openSUSE` here? :thinking: since I think should just work

README.md
```diff
@@ -2,48 +2,107 @@
...
-- show DRBD metrics (local and remote disks resource metrics)
+```
+systemctl --now enable prometheus-ha_cluster_exporter
+```
 
-We mantain a complete list of the [metric specification](doc/metric_spec.md), usage and possible values.
+## Development
 
-## Devel:
+This is a young project, pull requests are more than welcome!
```
MalloZup: I'm not really in favor to keep this. This is a young project,
I mean from experience in future we will forget to remove this string and it will stay forever young .. I wish also :grin:

MalloZup:
@stefanotorresi  a part of that, looks good to me. I would try to squash some commits that would be really appreciated . :cupid: :grin:

README.md
stefanotorresi: I thought about this, but we have never actually tested the project on openSUSE.

MalloZup: we can safe put openSUSE here.

stefanotorresi:
@MalloZup rebased

PR #92, opened 2019-11-11
Add SBD devices count metric

### LINKED ISSUE #90: Create SBD metric about number of device configured
the metric should give 0 if there is not sbd device is configured and give a number int of the number

stefanotorresi:
closes #90

### DIFF
--- a/doc/metric_spec.md
+++ b/doc/metric_spec.md
@@ -161,6 +161,7 @@ The SBD subsystems collect devices stats by parsing its configuration the output
 
 0. [Sample](../test/sbd.metrics)
 1. [`ha_cluster_sbd_device_status`](#ha_cluster_sbd_device_status)
+2. [`ha_cluster_sbd_devices_total`](#ha_cluster_sbd_devices_total)
 
 ### `ha_cluster_sbd_device_status`
 
@@ -175,6 +176,12 @@ Value is either `1` or `0`.
 
 The total number of lines for this metric will be the cardinality of `device`.
 
+
+### `ha_cluster_sbd_devices_total`
+
+#### Description
+
+Total count of configured SBD devices.
 
 ## DRBD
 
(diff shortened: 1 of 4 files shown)


doc/metric_spec.md
```diff
@@ -176,6 +177,12 @@ Value is either `1` or `0`.
 The total number of lines for this metric will be the cardinality of `device`.
 
 
+### `ha_cluster_sbd_devices_total`
+
+#### Description
+
+Total count of configured SBD devices.
```
MalloZup: `nit`

we should add here also what is the value it returns since most of metric provide it. ( it something imho we are not doing regularly :grin: )

But we should say:

Total count of cofingured SBD devices. Possible values from 0 to N integer or something perhaps.

sbd_metrics.go
```diff
@@ -96,27 +89,24 @@ func readSdbFile() ([]byte, error) {
...
 
 // retrieve a list of sbd devices from the config file contents
-func getSbdDevices(sbdConfigRaw []byte) ([]string, error) {
+func getSbdDevices(sbdConfigRaw []byte) []string {
 	// in config it can be both SBD_DEVICE="/dev/foo" or SBD_DEVICE=/dev/foo;/dev/bro
-	wordOnly := regexp.MustCompile("SBD_DEVICE=\"?[a-zA-Z-/;]+\"?")
-	sbdDevicesConfig := wordOnly.FindString(string(sbdConfigRaw))
-
-	// check the case there is an sbd_config but the SBD_DEVICE is not set
+	regex := regexp.MustCompile(`(?m)^\s*SBD_DEVICE="?((?:[\w-/];?)+)"?\s*$`)
```
MalloZup: uh this regex looks complicated. Why did we changed it? perhaps we can add a comment of what he is doing. The previous one was simpler.

stefanotorresi: Yeah, some commenting was indeed due!

doc/metric_spec.md
stefanotorresi: good idea. We should also do this for all the other ones, because actually non of the `*_total` metrics document their value.

stefanotorresi:
@MalloZup rebased!

PR #94, opened 2019-11-14
Fix matching labels and case sensitivity

stefanotorresi:
While using the metrics in Grafana I realized that using the common convention of differentiating discriminator fields between "owning" and "joined" side, which I borrowed from RDBMS, doesn't really work with Prometheus time series.

One example of this was 
````
ha_cluster_pacemaker_resources{id="foo"}
ha_cluster_pacemaker_fail_count{resource="foo"}
````
Originally, it was intentional to use distinct labels, because I thought Prometheus would allow specifying the label names to match in aggregated queries using `on` `group_*` and `by`, but I have only now found out that this is not the case: the label are actually assumed to have the same name.  
Grafana also has a few functionalities that work under this assumption.

This PR also fixes a bug where the resource name was mistakenly processed with `strings.ToLower`, which would also produce unexpected results with the aforementioned kind of queries.

The presented changeset constitutes a definite breaking change, that will need to be properly disclosed during the release.

### DIFF
--- a/drbd_metrics.go
+++ b/drbd_metrics.go
@@ -34,7 +34,7 @@ var (
 	drbdMetrics = metricDescriptors{
 		// the map key will function as an identifier of the metric throughout the rest of the code;
 		// it is arbitrary, but by convention we use the actual metric name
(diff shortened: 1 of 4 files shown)


MalloZup:
check out if we don't break any new dashboard or alerts with this :)

PR #99, opened 2019-11-19
Refactor resource location constraints metric

stefanotorresi:
This PR performs some changes to `ha_cluster_pacemaker_constraints` as some limitations have arisen during the implementation of its visual representation in the Grafana dashboard:

- **BC break:**  renames `ha_cluster_pacemaker_constraints` to `ha_cluster_pacemaker_location_constraints` to distinguish it from possible future new types of constraint metrics like order and colocation.
- **BC break:** the value is no longer always `1` but is now set to the _score_ of the constraint.
- **BC break:** the`type` label, which was not even documented before, has been removed.
- the `id` label has been renamed to `constraint`, to conform to the convention consolidated in #94.
- the new labels `node` and `role` have been added.
- all location constraints are scraped, not just the ones with their id starting with `cli-ban` or `cli-prefer`. That naming prefix was a convention of `crmsh`, but users may be adding constraints through other means (e.g. cibadmin, hawk, crm_resource, etc).
- the `cibadmin` command no longer uses an `--xpath` flag that would produce non-0 exit codes in case the XML query wouldn't return any result, causing the whole pacemaker collector to go into a warning condition.
- The whole CIB is now unmarshalled in a more generic data structure that can accommodate for further metrics.
- `fake_cibadmin.sh` now produces a complete CIB XML mock.
- metric documentation was outdated and has been updated.

### DIFF
--- a/doc/metric_spec.md
+++ b/doc/metric_spec.md
@@ -25,8 +25,8 @@ The Pacemaker subsystem collects an atomic snapshot of the HA cluster directly f
 
 0. [Sample](../test/pacemaker.metrics)
 1. [`ha_cluster_pacemaker_config_last_change`](#ha_cluster_pacemaker_config_last_change)
-2. [`ha_cluster_pacemaker_constraints`](#ha_cluster_pacemaker_constraints)
 3. [`ha_cluster_pacemaker_fail_count`](#ha_cluster_pacemaker_fail_count)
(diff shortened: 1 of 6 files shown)


MalloZup:
@stefanotorresi  a part of that, squash the last commit `fmt fix`

PR #100, opened 2019-11-21
Add split brain metric 

MalloZup:
# Description:

this pr add splitbrain metric in conjoction with our handler.

# WHat is left:

before we merge this I would like to test the exporter withing the real splitbrain but sofar I tested manually everything worked.

# Info:

I just removed the label `node` because we don't needed it. This is already provided by `instance` proemtheus label.


doc/metric_spec.md
```diff
@@ -249,3 +249,42 @@ Either the value is `1`, or the line is absent altogether.
...
+1) create hook:
+
+```bash
+#! /bin/bash
+mkdir -p /var/run/drbd/splitbrain
+echo "DRBD split-brain detected! Remember to remove the file /var/lib/drbd/drbd-split-brain-detected once the splitbrain is solved!" > /var/run/drbd/splitbrain/drbd-split-brain-detected-$DRBD_RESOURCE-$DRBD_VOLUME
+```
+
+2) copy this hook into all drbd nodes:
+```/var/lib/drbd/notify-split-brain-haclusterexporter-suse-metric.sh```
```
stefanotorresi: rather than being so explicit and risking of repeating ourselves with the other project, why don't we refer to it directly?

e.g.
```
1) get the hook from [SUSE/ha-sap-terraform-deployments](https://github.com/SUSE/ha-sap-terraform-deployments/blob/88a2b5612e92d34976c53628ba9f33300a63bd8b/salt/drbd_node/files/notify-split-brain-haclusterexporter-suse-metric.sh)
```

doc/metric_spec.md
```diff
@@ -249,3 +249,42 @@ Either the value is `1`, or the line is absent altogether.
...
+### `ha_cluster_drbd_split_brain`
+
+#### Description
+
+This metric signal if there is a split brain occuring per resource and volume.
+Either the value is `1`, or the line is absent altogether.
+
+This metric is a special metric comparing to others, because in order to make this metric working you will need to set a drbd customer split-brain handler. Look at the end 
+
+### Labels
```
stefanotorresi: Heading hierarchy is off

```suggestion
#### Labels
```

doc/metric_spec.md
```diff
@@ -249,3 +249,42 @@ Either the value is `1`, or the line is absent altogether.
...
+
+- `resource`: the name of the resource.
+- `volume`: the volume number
+
+### Setting up the DRBD split brain hook:
```
stefanotorresi: Heading hierarchy is off

```suggestion
#### Setting up the DRBD split-brain hook
```

doc/metric_spec.md
```diff
@@ -249,3 +249,42 @@ Either the value is `1`, or the line is absent altogether.
...
+
+In order to get woring the split_brain metric:
```
stefanotorresi:
```suggestion
In order to get the `split_brain` metric working:
```

drbd_metrics.go
```diff
@@ -37,6 +39,7 @@ var (
 		"resources":        NewMetricDesc("drbd", "resources", "The DRBD resources; 1 line per name, per volume", []string{"resource", "role", "volume", "disk_state"}),
 		"connections":      NewMetricDesc("drbd", "connections", "The DRBD resource connections; 1 line per per resource, per peer_node_id", []string{"resource", "peer_node_id", "peer_role", "volume", "peer_disk_state"}),
 		"connections_sync": NewMetricDesc("drbd", "connections_sync", "The in sync percentage value for DRBD resource connections", []string{"resource", "peer_node_id", "volume"}),
+		"split_brain":      NewMetricDesc("drbd", "split_brain", "The split brain metric per node, resource and volume. The value 1 indicate the split brain is present", []string{"resource", "volume"}),
```
stefanotorresi: I would propose some rewording of the help message:

```suggestion
		"split_brain":      NewMetricDesc("drbd", "split_brain", "Whether a split brain has been detected; 1 line per resource, per volume.", []string{"resource", "volume"}),
```

doc/metric_spec.md
```diff
@@ -249,3 +249,42 @@ Either the value is `1`, or the line is absent altogether.
...
+```
+
+2) copy this hook into all drbd nodes:
+```/var/lib/drbd/notify-split-brain-haclusterexporter-suse-metric.sh```
+
+3) on the drbd configuration enable the hook:
+
+```split_brain: "/usr/lib/drbd/notify-split-brain-haclusterexporter-suse-metric.sh"`
+
+Refer to upstream doc: https://docs.linbit.com/docs/users-guide-8.4/#s-configure-split-brain-behavior
```
stefanotorresi: Maybe it's worth mentioning the full solution documentation (https://github.com/SUSE/ha-sap-terraform-deployments/blob/develop/doc/monitoring.md#drbd-split-brain-detection) so that users understand they don't need to do any of these steps if they're using the exporter within the full solution.

drbd_metrics.go
```diff
@@ -65,6 +68,28 @@ func (c *drbdCollector) Collect(ch chan<- prometheus.Metric) {
...
+				if len(strings.Split(resAndVolume, "-")) == 2 {
+					resource := strings.Split(resAndVolume, "-")[0]
+					volume := strings.Split(resAndVolume, "-")[1]
+					ch <- c.makeGaugeMetric("split_brain", float64(1), resource, volume)
+					log.Warnf("Found split brain for drbd %s %s! Please remove file once you have resolved splitbrain", resource, volume)
+				}
+			}
+		}
+	}
+
```
stefanotorresi: So far, so good! Though, there are still a few things that I'd like to see improved in the implementation:

1. I think the directory should be configurable and not hard-coded! I really don't like hard-coded stuff, and now that we have a configuration system in place we have no excuses... ;)
2. The `Collect` one has gotten quite big. I think we should refactor it like the pacemaker collector, and have the two main collection blocks separated, e.g. `c.recordResourceMetrics` and `c.recordSplitBrainMetric`.
3. Rather than nesting all these `if` blocks one into another, consider the  "fail fast, fail early" approach: that would make the most relevant part of the code stand out, instead of being hidden by multiple levels of indentation.

MalloZup:
ok thx I will adress comments next iteration :package:

doc/metric_spec.md
MalloZup: @stefanotorresi  I tought about this. On the other side I didn't want to have a broken link :grin:  if we move location. Both approaches have downsides imho.

I think having redundant data,  can be safer to broken link. But I can change it  :grin:

doc/metric_spec.md
MalloZup: ok

drbd_metrics.go
MalloZup: Yep I agree. The thing about 3) is that I might need to

THe thing is that the drbd collector has 2 sources of inputs. ( the binary drbd

doc/metric_spec.md
MalloZup: we can have the links and see how it goes :grin:

drbd_metrics.go
MalloZup: done

MalloZup:
@stefanotorresi even if you approve, don't merge. I would like to test with some custom script provided by Nick etc. I have already tested it with injecting files etc it works, but I want to test the whole solution once more or less.

drbd_metrics.go
```diff
@@ -111,3 +120,28 @@ func parseDrbdStatus(statusRaw []byte) ([]drbdStatus, error) {
...
 }
+
+func (c *drbdCollector) setDrbdSplitBrainMetric(ch chan<- prometheus.Metric) {
+
+	// set split brain metric
+	// by default if the custom hook is not set, the exporter will not be able to detect it
+	files, err := ioutil.ReadDir(c.drbdSplitBrainPath)
+	if err != nil {
+		log.Warnf("Error while reading directory /var/run/drbd/splitbrain: %s", err)
+	}
```
stefanotorresi: You probably forgot to use the `c.drbdSplitBrainPath` variable in the log message!

drbd_metrics.go
```diff
@@ -111,3 +120,28 @@ func parseDrbdStatus(statusRaw []byte) ([]drbdStatus, error) {
...
+		if match != nil {
+			resAndVolume := strings.Split(f.Name(), "drbd-split-brain-detected-")[1]
+			// avoid to have index out range panic error (in case the there is not resource-volume syntax)
+			if len(strings.Split(resAndVolume, "-")) == 2 {
+				resource := strings.Split(resAndVolume, "-")[0]
+				volume := strings.Split(resAndVolume, "-")[1]
+				ch <- c.makeGaugeMetric("split_brain", float64(1), resource, volume)
+				log.Warnf("Found split brain for drbd %s %s! Please remove file once you have resolved splitbrain", resource, volume)
+			}
+		}
```
stefanotorresi: Here is what I meant by "fail fast, fail early":

Instead of:
```go
if match != nil {
   // happy path
}
```
consider:
```go
if match == nil {
    continue
}
// happy path
```

Same goes for the length check in the string split.
Talking about the string split, we are wastefully calling the same split over and over: I think we should rather store the result of the split in a variable:

```go
split := strings.Split(resAndVolume, "-")
if len(split) != 2 {
    // maybe log a warning
    continue
}
// etc
```

I also don't think we should log a warning about the system being monitored, we should only log the status of the exporter itself!

doc/metric_spec.md
stefanotorresi: if we use the github permaurl of the file, it will never be broken ;)

drbd_metrics_test.go
```diff
@@ -181,6 +181,6 @@ func TestNewDrbdCollectorChecksDrbdsetupExecutableBits(t *testing.T) {
 func TestDRBDCollector(t *testing.T) {
 	clock = StoppedClock{}
 
-	collector, _ := NewDrbdCollector("test/fake_drbdsetup.sh")
+	collector, _ := NewDrbdCollector("test/fake_drbdsetup.sh", "splitbrainfake")
 	expectMetrics(t, collector, "drbd.metrics")
 }
```
stefanotorresi: what about adding a dedicated test case? it should be easy enough now that the dedicated method is extracted.

drbd_metrics.go
MalloZup: make sense done

doc/metric_spec.md
MalloZup: done

drbd_metrics_test.go
MalloZup: unfortunately isn't that easy enough :grin: 

Since the metrics aren't data but directory/files we need to create and remove.

MalloZup: done

MalloZup:
thx. I will link to this PR the other one. I will let this pr a little pending until tomorrow since I want to test the hook etc.

MalloZup:
related : https://github.com/SUSE/ha-sap-terraform-deployments/pull/277

Issue #97, opened 2019-11-15
ha_cluster_exporter - provide different default configuration for different versions of the OS


MalloZup:
in the OBS build, we should bundle a different configuration file in `/usr/etc` for the versions of SLE that require so, so that the safe default paths for the external tools are correctly set and the package always works OOTB.

PR #101, opened 2019-11-22
change default drbdsetup-path config value

stefanotorresi:
Following my attempt at delivering different configurations for different distributions, I found out that none of the supported ones actually deviates from the defaults; the original cause of #97 was that `drbdsetup` is always installed in `/sbin` rather than `/usr/sbin`.


stefanotorresi:
Thanks for the encouragement @diegoakechi! We will certainly make good use of it in the future. :)

PR #102, opened 2019-11-27
OBS automated releases

stefanotorresi:
This PR adds automated releases to OBS and automatic assets uploads in GH releases.

To perform a release, just create a release/tag in GH, or even push a tag from a local clone.

This required setting some environment variables in Travis, which I already have taken care of.

It is also possible to produce the obs working directory locally by running `VERSION=$TAG make obs-workdir`, which will prepare everything in the temporary directory `build/obs`, and then commit to OBS with `make obs-commit`.
These make targets assume a configured `osc` host, which in Travis is provided with Docker.


Makefile
```diff
@@ -1,13 +1,22 @@
-default: clean static-checks test build post-build
+VERSION ?= dev
```
MalloZup: comment to this variable would not harm in context of their behaviour perhaps

stefanotorresi: @MalloZup good call! I added some comments for all the Makefile variables.

stefanotorresi:
@arbulu89 

> So now, a new release will be created with every new tag, right?

Yes, although you need a GH release to produce a proper new entry in the RPM changelog.

> Any example of the created `changes` file?

if I run this
```shell
for TAG in $(git tag); do 
    .ci/gh_release_to_obs_changeset.py clusterlabs/ha_cluster_exporter \
        --author "shap-staff@suse.de"
        --tag $TAG 
        -f build/obs/prometheus-ha_cluster_exporter.changes
done
```

here is the produced [changelog](https://github.com/ClusterLabs/ha_cluster_exporter/files/3901602/changelog.txt)

MalloZup:
@stefanotorresi  why we merged this? I was not even done with review :(

MalloZup:
I will check perhaps we can do some follow up

MalloZup:
@stefanotorresi  @arbulu89  we need to be more strict about commit message. 

I think I'm flexibile but commit message with dummynames we need to remove them in future.

PR #105, opened 2019-12-03
Move the build back to obs

stefanotorresi:
This PR slightly changes the continuous deployment process to revert building the binaries bundled in RPM packages inside SUSE's OBS system, due to compliance requirements.

OBS project repos should be checked to work for the supported version once this gets released.


stefanotorresi:
I knew you were right!! I just wanted to try and do without that OBS abomination. 😛

PR #106, opened 2019-12-04
Add more DRBD resource metrics

kermat:
This pull request adds more of DRBD's metrics and the respective tests to the exporter.

### DIFF
--- a/drbd_metrics.go
+++ b/drbd_metrics.go
@@ -19,6 +19,12 @@ type drbdStatus struct {
 	Role    string `json:"role"`
 	Devices []struct {
 		Volume    int    `json:"volume"`
+		Written   int    `json:"written"`
+		Read      int    `json:"read"`
+		AlWrites  int    `json:"al-writes"`
+		BmWrites  int    `json:"bm-writes"`
+		UpPending int    `json:"upper-pending"`
+		LoPending int    `json:"lower-pending"`
 		DiskState string `json:"disk-state"`
 	} `json:"devices"`
 	Connections []struct {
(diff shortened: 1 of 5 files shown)


drbd_metrics.go
```diff
@@ -36,10 +46,20 @@ var (
...
+		"resources":            NewMetricDesc("drbd", "resources", "The DRBD resources; 1 line per name, per volume", []string{"resource", "role", "volume", "disk_state"}),
+		"written":              NewMetricDesc("drbd", "written", "KiB written to DRBD; 1 line per res, per volume", []string{"resource", "role", "volume", "disk_state"}),
+		"read":                 NewMetricDesc("drbd", "read", "KiB read from DRBD; 1 line per res, per volume", []string{"resource", "role", "volume", "disk_state"}),
+		"al_writes":            NewMetricDesc("drbd", "al_writes", "Writes to activity log; 1 line per res, per volume", []string{"resource", "role", "volume", "disk_state"}),
+		"bm_writes":            NewMetricDesc("drbd", "bm_writes", "Writes to bitmap; 1 line per res, per volume", []string{"resource", "role", "volume", "disk_state"}),
+		"upper_pending":        NewMetricDesc("drbd", "upper_pending", "Upper pending; 1 line per res, per volume", []string{"resource", "role", "volume", "disk_state"}),
+		"lower_pending":        NewMetricDesc("drbd", "lower_pending", "Lower pending; 1 line per res, per volume", []string{"resource", "role", "volume", "disk_state"}),
```
stefanotorresi: I'm not sure we need `disk_state` and `role` labels here, what do you think about removing them and leaving just `volume` and `resource`?

kermat: You're right, I'll drop those labels.

kermat: And, I'll update the docs while I'm at it :)

kermat: Done and pushed! Thanks @stefanotorresi

PR #108, opened 2019-12-07
Add quorum status to DRBD metrics

kermat:
Hello Again!

I wanted to quickly add one more DRBD metric, quorum status, to the exporter. Sorry for not including it in the last PR.

drbd_metrics.go: Added quorum boolean to metrics.
drbd_metrics_test.go: Added test for quorum boolean and added some
  missing tests (missing from PR#106) for second resource.
test/drbd.metrics: added appropriate test values for new metric.
doc/metric_spec.md: added documentation for new metric.

### DIFF
--- a/drbd_metrics_test.go
+++ b/drbd_metrics_test.go
@@ -178,15 +178,7 @@ func TestDrbdParsing(t *testing.T) {
 		t.Errorf("received should be 456")
 	}
 
-	if 456 != drbdDevs[1].Connections[0].PeerDevices[0].Received {
-		t.Errorf("received should be 456")
-	}
-
 	if 654 != drbdDevs[0].Connections[0].PeerDevices[0].Sent {
-		t.Errorf("sent should be 654")
-	}
-
-	if 654 != drbdDevs[1].Connections[0].PeerDevices[0].Sent {
 		t.Errorf("sent should be 654")
 	}
 
@@ -194,15 +186,7 @@ func TestDrbdParsing(t *testing.T) {
 		t.Errorf("pending should be 3")
 	}
 
-	if 3 != drbdDevs[1].Connections[0].PeerDevices[0].Pending {
-		t.Errorf("pending should be 3")
-	}
-
 	if 4 != drbdDevs[0].Connections[0].PeerDevices[0].Unacked {
-		t.Errorf("unacked should be 4")
-	}
-
-	if 4 != drbdDevs[1].Connections[0].PeerDevices[0].Unacked {
 		t.Errorf("unacked should be 4")
 	}
 
--- a/drbd_metrics_test.go
+++ b/drbd_metrics_test.go
@@ -71,7 +71,7 @@ func TestDrbdParsing(t *testing.T) {
         "minor": 3,
         "disk-state": "UpToDate",
         "client": false,
-        "quorum": true,
+        "quorum": false,
         "size": 10200,
         "read": 654321,
         "written": 123456,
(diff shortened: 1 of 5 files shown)


drbd_metrics_test.go
```diff
@@ -166,22 +166,46 @@ func TestDrbdParsing(t *testing.T) {
 		t.Errorf("lower-pending should be 2")
 	}
 
+	if true != drbdDevs[0].Devices[0].Quorum {
+		t.Errorf("quorum should be true")
+	}
+
+	if true != drbdDevs[1].Devices[0].Quorum {
```
diegoakechi: I would suggest you change the fake drbd data to have devices with and without quorum, so we test all possibilities here

drbd_metrics_test.go
```diff
@@ -166,22 +166,46 @@ func TestDrbdParsing(t *testing.T) {
...
+		t.Errorf("quorum should be true")
+	}
+
 	if 456 != drbdDevs[0].Connections[0].PeerDevices[0].Received {
 		t.Errorf("received should be 456")
 	}
 
+	if 456 != drbdDevs[1].Connections[0].PeerDevices[0].Received {
```
diegoakechi: Can you separate the UT from the previous PR in separated PR, or at least a new commit? It helps to better organize the changes and the repo history.

drbd_metrics_test.go
kermat: Good call, will do!

drbd_metrics_test.go
kermat: Sure thing! Thanks for the quick feedback.

drbd_metrics_test.go
kermat: pushed new commit setting the second resource and expected results to `false`. Thanks!

drbd_metrics_test.go
kermat: Removed the tests in a subsequent commit. I'll issue another PR to fix these up later on. Thanks again!

test/drbd.metrics
```diff
@@ -46,6 +46,10 @@ ha_cluster_drbd_upper_pending{resource="1-single-1",volume="0"} 1 1234
 # TYPE ha_cluster_drbd_lower_pending gauge
 ha_cluster_drbd_lower_pending{resource="1-single-0",volume="0"} 2 1234
 ha_cluster_drbd_lower_pending{resource="1-single-1",volume="0"} 2 1234
+# HELP ha_cluster_drbd_quorum Quorum status; 1 line per res, per volume
```
MalloZup: @kermat  a `nit`: 

I think we could definitely improve the `help` description message perhaps.

I mean 1 line doesn't in which context? 

We could imho remove the 1 line. ( this would apply also for other metrics)

` HELP ha_cluster_drbd_quorum Quorum status per resource and per volume`
	
Something in this direction

drbd_metrics_test.go
MalloZup: thx

drbd_metrics_test.go
MalloZup: good point thx

test/drbd.metrics
kermat: I agree. I was going to add some missing tests for the second fake resource in a separate PR to keep things tidy per diegoakechi's recommendation. So, I'll change the quorum description in this PR, and open another PR to address the missing tests and messages added in PR #106

kermat:
I've pushed the suggested changes to the `help` description.

MalloZup:
Will look tomorrow thx sofar!

PR #110, opened 2019-12-12
Remove unrelevant warning for splitbrain file

### LINKED ISSUE #107: Does not report error if the DRBD split brain file does not exist
The exporter is reporting error because the split-brain directory does not exist

```
Dec 06 13:14:11 vmdrbd01 ha_cluster_exporter[32008]: time="2019-12-06T13:14:11Z" level=warning msg="Error while reading directory /var/run/drbd/splitbrain: open /var/run/drbd/splitbrain: no such file or directory"
```

As this means that there is no split-brain, it should not be reported. If the file exists, this should be logged and should also explain that the sysadmin needs to remove the file after threating the split-brain.

MalloZup:
# What does this PR:

this fix #107 

On the otherside, this PR has the minimalist approach. 

I have been thought a while if we should report the existence of the file with a warn. But I think we can skip it. Because a part of the technical complexity (not major but kind of) to create a singleton object on the loop iteration (if we don't do it we will polluate logs with all this entry),

I think we can avoid to create the entry, since the user should know that if the exporter say that the metric is set to 1, he should remove the file. this is something we say also explitely on the metric usage and doc.

### DIFF
--- a/drbd_metrics.go
+++ b/drbd_metrics.go
@@ -173,11 +173,9 @@ func (c *drbdCollector) setDrbdSplitBrainMetric(ch chan<- prometheus.Metric) {
 
 	// set split brain metric
 	// by default if the custom hook is not set, the exporter will not be able to detect it
-	files, err := ioutil.ReadDir(c.drbdSplitBrainPath)
-	if err != nil {
-		log.Warnf("Error while reading directory %s: %s", c.drbdSplitBrainPath, err)
-	}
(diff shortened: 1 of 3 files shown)


stefanotorresi:
@MalloZup do we have a test case covering the fact that the directory we try to browse doesn't exist?

MalloZup:
opened follow-up issue https://github.com/ClusterLabs/ha_cluster_exporter/issues/112

PR #111, opened 2019-12-12
[WIP]: Improve drbd metric help message

### LINKED ISSUE #109: Drbd metric help-message need some improvemnt
we should polish the `help` message of some DRBD metrics.
https://github.com/ClusterLabs/ha_cluster_exporter/blob/master/drbd_metrics.go#L49

We don't need the per line text etc. see pr #108

MalloZup:
fix #109 

This PR doesn't do much

IT just improve the HELP message of the metrics of the drbd which was not coherent along all metrics . Now is better

### DIFF
--- a/drbd_metrics.go
+++ b/drbd_metrics.go
@@ -47,15 +47,15 @@ var (
 	drbdMetrics = metricDescriptors{
 		// the map key will function as an identifier of the metric throughout the rest of the code;
 		// it is arbitrary, but by convention we use the actual metric name
-		"resources":            NewMetricDesc("drbd", "resources", "The DRBD resources; 1 line per name, per volume", []string{"resource", "role", "volume", "disk_state"}),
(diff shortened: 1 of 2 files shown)


stefanotorresi:
Erm... I find the new lines to be more confusing than before, to be honest. :nerd_face: 
1. The semicolon was there to separate the description of the metric from the indication of the cardinality.
2. I don't see the point of using the "metric" word, because of course these are all metrics.
3. There are a lot of double white spaces which hinder readability.

MalloZup:
@stefanotorresi  I personally don't like the `1 line per res` text.

To me 1) res should be resource
2) and 1line per res  is somehow confusing.
So imho we could remove that  

the rest can be changed no pb

stefanotorresi:
@MalloZup I'm fine with completely removing the cardinality indication, and leave just the description! ;)

PR #116, opened 2019-12-28
Include test files in release

vvidic:
Debian packaging of the exporter uses release .tar.gz files and it would be good to run tests.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,2 @@
 /.* export-ignore
-/test export-ignore
-*_test.go export-ignore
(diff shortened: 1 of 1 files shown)


diegoakechi:
@MalloZup it shouldn't be a problem for OBS, and actually it is a good practice to include the test in our build system too, so it still runs when someone eventually forks the package inside OBS and perform some change.

MalloZup:
ok thx

MalloZup:
thx @vvidic for this

PR #123, opened 2020-02-05
Refactor metrics

stefanotorresi:
This PR addresses some of the feedback we got in prometheus/docs#1540, which is part of the submission of this package in the [curated list of third party exporters](https://prometheus.io/docs/instrumenting/exporters/).

- **BC Break**: The following metrics have been removed, because they were mostly redundant, and their name broke Prometheus conventions:
  - `ha_cluster_pacemaker_nodes_total`
  - `ha_cluster_pacemaker_resources_total`
  - `ha_cluster_sbd_devices_total`
- **BC Break**: The `ha_cluster_corosync_ring_errors_total` has been renamed to `ha_cluster_corosync_ring_errors` to comply with Prometheus naming conventions.
- **BC Break**: The `ha_cluster_sbd_device_status` has been renamed to `ha_cluster_sbd_devices`, which now works similarly to other metrics that represent entities in the cluster: there will be one line per SBD device valued `1`, and the status is a label.
- refactored the creation of the metric descriptors in a `setDescriptor()` function targeting the `DefaultCollector` struct. This got rid of some global state and a whole lot of repetition and ambiguity in the previous metric description instantiation.
- the metric subsystem prefix is now initialized just once, within the collector structs.

Any PromQL query that took advantage of the removed `*_total` metrics can be easily adapted by removing the `_total` suffix from the metric name and using the `count()` function.

While we understand that these changes involve some backwards compatibility breakages that may result annoying to our users, we feel that, as we approach our first 1.0 stable release, addressing the feedback we got from the Prometheus maintainers is very important for the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ go get github.com/ClusterLabs/ha_cluster_exporter
 ```
 
 ### RPM
-You can find the repositories for RPM based distributions in [SUSE's Open Build Service](https://build.opensuse.org/repositories/server:monitoring/prometheus-ha_cluster_exporter).  
+You can find the repositories for RPM based distributions in [SUSE's Open Build Service](https://build.opensuse.org/package/show/server:monitoring/prometheus-ha_cluster_exporter).  
 On openSUSE or SUSE Linux Enterprise you can just use the `zypper` system package manager:
 ```shell
 export DISTRO=SLE_15_SP1 # change as desired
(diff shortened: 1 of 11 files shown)


stefanotorresi:
come on man, fmt it's just one commit, squashing would take more time than it took to make it! :stuck_out_tongue:

diegoakechi:
@stefanotorresi This pretty much broke some of the panels on the Cluster dashboards on https://github.com/SUSE/ha-sap-terraform-deployments. Please, consider to fix it ASAP. 

Thanks.

Issue #114, opened 2019-12-19
`ha_cluster_pacemaker_resources` doesn't report stopped resources


stefanotorresi:
Steps to reproduce:

1. log into a pacemaker node with the exporter running
2. add a dummy resource with `crm configure primitive test Dummy`
3. the query `ha_cluster_pacemaker_resources{resource="test"}` will return 1 line with `role="started"`
4. stop the resource with `crm resource stop test`
5. wait 5 mins for prometheus metric staleness (related #113)
6. query `ha_cluster_pacemaker_resources{resource="test"}` 

expected behaviour:
the query should return 1 line with `role="stopped"`

observed behaviour:
the query returns 0 results

stefanotorresi:
This is even worse than initial thought, as it doesn't only affect primitive resources, but grouped ones:
I just tried to stop a resource which was part of a group, and since we're using the `--group-by-nodes` flag with `crm_mon`, the stopped resource no longer appears as a `<node>` XML child, so we don't export it.

Fixing this issue will probably need some non-trivial work on how we use the XML.

PR #132, opened 2020-02-20
Refactor Pacemaker collector

stefanotorresi:
This is a 100% transparent refactoring, no functionality is changed, but it's propedeutic for fixing #114.

The Pacemaker collector is now stripped from the responsibility of parsing the XML collected via `cibadmin` and `crmmon`, which is now done in dedicated packages, and exposed with a couple of dedicated interfaces (`cib.Parser` and `crmmon.Parser`), with dedicated XML unmarshalling structs which can now evolve independently without name clashes.


stefanotorresi:
@MalloZup good idea! I will update the design doc and ping you again.

PR #133, opened 2020-02-21
Clean up rpm spec and change license

stefanotorresi:
The build step was more complicated than necessary and the `License` field was wrong


stefanotorresi:
I'm considering if we should change the license instead, so don't merge this for now.

stefanotorresi:
I didn't remove it, I changed the license from GPLv3 to Apache 2.0.

PR #134, opened 2020-02-26
Fix tracking stopped resources

### LINKED ISSUE #114: `ha_cluster_pacemaker_resources` doesn't report stopped resources

stefanotorresi:
This PR fixes #114.
Only tested on primitive Pacemaker resources, tests on grouped ones will be performed at a later stage due to problems with the deployment repo.


collector/pacemaker/pacemaker.go
```diff
@@ -95,41 +96,55 @@ func (c *pacemakerCollector) recordNodeMetrics(crmMon crmmon.Root, ch chan<- pro
 			nodeType = "unknown"
 		}
 
-		for nodeStatus, isActive := range nodeStatuses {
-			if isActive {
+		for nodeStatus, flag := range nodeStatuses {
```
MalloZup: why we do rename it to flag?

MalloZup: I think `flag` is to generic and I don't get the context. Can we use a better var name?

stefanotorresi: I did this because it was confusing at first when reading this code again, due to the fact that we have the concept of "active" status, but `isActive` doesn't mean that, it means that a particular boolean value is true or not, so a more generic `flag` term is more appropriate.

MalloZup: ok still, I don't get what the flag is signaling :grin: .

I think we need a comment what the variable is doing.. I mean in which context to node is used. I'm talking without a memory of this piece of code.

stefanotorresi: good idea, I will add a comment explaining the flag looping, because if we forgot it already we'll certainly forget it again. :D

MalloZup: yeah indeed :+1:

stefanotorresi: @MalloZup I hope 9b8b8988f1df3ccdbddc10e16af2846c98934331 is clear enough, let me know if there is more we can do.

MalloZup: looks great now. I can understand with beeing ingnorant on code I wrote time ago. that is perfect :grin:

PR #135, opened 2020-02-28
Minor CI fixes

stefanotorresi:
There were 2 bugs in the CI setup:
1. In the `make obs-workdir` target, the `rm` command that cleans up existing archives could fail if no archive were present in the obs source tree. Adding a `-f` flag fixes this.
2. The build would not fail if the OBS deployment Docker command fails. To fix this, I used the suggestion from travis-ci/travis-ci#1574.
3. The python script converting GitHub releases to RPM changelogs would not create a changelog file if none was present.

### DIFF
--- a/.ci/gh_release_to_obs_changeset.py
+++ b/.ci/gh_release_to_obs_changeset.py
@@ -55,7 +55,7 @@ with tempfile.TemporaryFile("r+") as temp:
     temp.seek(0)
 
     if args.file:
-        with open(args.file, "r") as prev:
+        with open(args.file, "a+") as prev:
             old = prev.read()
         with open(args.file, "w") as new:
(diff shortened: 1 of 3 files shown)


stefanotorresi:
@arbulu89 all we need to do is these 8 lines of bash we have in the makefile:

https://github.com/ClusterLabs/ha_cluster_exporter/blob/a8dd60d91de04e7236fc7299514a6a920e6ad4e2/Makefile#L65-L76

yes, it's very similar to https://github.com/arbulu89/continuous-delivery/blob/b8ab10dc94bb011c128ada9878dc560e5500ccb8/scripts/upload.sh
but it's arguably much simpler, and there were a few details that I couldn't easily adapt in the existing script.

PR #142, opened 2020-03-19
unregister go_* metrics if loglevel is lower than debug

stefanotorresi:


ha_cluster_exporter.go
```diff
@@ -99,6 +99,10 @@ func main() {
 		log.Info("DRBD collector registered")
 	}
 
+	if !log.IsLevelEnabled(log.DebugLevel) {
```
MalloZup: can we add here a comment explaining what we do?

stefanotorresi: done!

PR #151, opened 2020-04-24
Corosync new features

### LINKED ISSUE #16: Add label RING_ID to corosync metric
Create a metric
`corosync_ring_errors("ring_id=1") 1
`corosync_ring_errors("ring_id=42") 0

 here so we can point what ring is failing.  A zero or 1 is a failure.

stefanotorresi:
This PR closes #16 by introducing a new `ha_cluster_corosync_rings` metric with the requested labels.

It also introduces `ha_cluster_corosync_member_votes`, which exposes how single nodes contribute to form a quorum.

The Corosync tools parsing has been heavily refactored in a similar way to what was done already for the Pacemaker, by introducing a `Parser` interface. This will allow us to more easily extend the support to future Corosync versions.


collector/corosync/parser.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func parseMembers(quorumToolOutput []byte) (members []Member, err error) {
+	sectionRE := regexp.MustCompile(`(?m)Membership information\n-+\s+Nodeid\s+Votes\s+Name\n((?:\w+\s+\d+\s[\w-]+(?:\s\(local\))?\n?)+)`)
+	sectionMatch := sectionRE.FindSubmatch(quorumToolOutput)
+	if sectionMatch == nil {
+		return nil, errors.New("could not find membership information")
+	}
+
+	linesRE := regexp.MustCompile(`(?m)(?P<node_id>\w+)\s+(?P<votes>\d+)\s(?P<name>[\w-]+)(?:\s(?P<local>\(local\)))?\n?`)
```
MalloZup, line 178: can we have maybe a comment before each regex to explain what we are looking for?

collector/corosync/parser.go
```diff
@@ -0,0 +1,213 @@
...
+	quorumVotes.Quorum, err = strconv.ParseUint(string(matches[4]), 10, 64)
+	if err != nil {
+		return quorumVotes, errors.Wrap(err, "could not parse vote number to uint64")
+	}
+
+	return quorumVotes, nil
+}
+
+func parseMembers(quorumToolOutput []byte) (members []Member, err error) {
+	sectionRE := regexp.MustCompile(`(?m)Membership information\n-+\s+Nodeid\s+Votes\s+Name\n((?:\w+\s+\d+\s[\w-]+(?:\s\(local\))?\n?)+)`)
```
MalloZup, line 172: can we have comment highlevel to explain regex what we want? so it can improve regex maintanance which as usual you remeber for 1 day and then forget

collector/corosync/parser.go
```diff
@@ -0,0 +1,213 @@
...
+
+func parseQuoromVotes(quorumToolOutput []byte) (quorumVotes QuorumVotes, err error) {
+	re := regexp.MustCompile(`(?m)Expected votes:\s+(\d+)\s+Highest expected:\s+(\d+)\s+Total votes:\s+(\d+)\s+Quorum:\s+(\d+)`)
+
+	matches := re.FindSubmatch(quorumToolOutput)
+	if matches == nil {
+		return quorumVotes, errors.New("could not find quorum votes numbers")
+	}
+
+	quorumVotes.ExpectedVotes, err = strconv.ParseUint(string(matches[1]), 10, 64)
```
MalloZup, line 148: are we sure here we don't get a `panic: runtime error: index out of range` ?
In case match length is 0 only

collector/corosync/corosync.go
```diff
@@ -36,129 +35,70 @@ type corosyncCollector struct {
...
 }
 
 func (c *corosyncCollector) Collect(ch chan<- prometheus.Metric) {
 	log.Debugln("Collecting corosync metrics...")
 
-	err := c.collectRingErrorsTotal(ch)
-	if err != nil {
-		log.Warnf("Corosync Collector scrape failed: %s", err)
-	}
+	// We suppress the exec errors because if any interface is faulty the tools will exit with code 1, but we still want to parse the output.
```
MalloZup, line 44: why we need to parse a broken output instead of returning error before?

collector/corosync/parser.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+type Status struct {
+	NodeId      string
+	RingId      string
+	Seq         uint64
+	Rings       []Ring
+	QuorumVotes QuorumVotes
+	Quorate     bool
+	Members     []Member
```
MalloZup, line 22: my personal pov is having nested struct looks more compact rather then all this sparsed structs . but is more taste. I prefer however compacted struct they are easier to read, instead of jumping around to mini-struct tied together.

collector/corosync/corosync.go
stefanotorresi: this logic was already present: we want to still parse the output in case of failure because we need to report faulty rings.

collector/corosync/parser.go
stefanotorresi: nope: since we have a mandatory capture group in the regex we will always have two matches in the result.
If it doesn't match at all, then we guard this in the two lines above.

collector/corosync/parser.go
stefanotorresi: yeah, good call.

collector/corosync/parser.go
stefanotorresi: yeah, that's definitely very subjective: I personally don't like nested structures, as you can't type hint to them individually.

collector/corosync/corosync.go
MalloZup: ok

collector/corosync/parser.go
MalloZup: ok it was just I wanted to highlight in sake, I will do a pr when you sleep
so I can change it again.. Joking :grin:

stefanotorresi:
@MalloZup I added some docs, let me know if you find them explanatory enough.

PR #152, opened 2020-04-27
avoid disappearing metrics for pacemaker nodes and resources

stefanotorresi:
This PR implements what we discussed at the end last week, following up some feedback we received from MSFT.

`ha_cluster_pacemaker_nodes` and `ha_cluster_pacemaker_resources` now will report a `0` value instead of a missing metric line for those status flags that are not `true` in the CIB.
This way, we correctly report all the possible combinations, including the ones where all the statuses are off. This will result in a more verbose metric set, but it's much clearer in intent and usable from a consumption perspective.

This PR also removes:
- the previously introduced `unknown` node type, which is not needed because the upstream data source has `member` as a default value.
- the artificial empty resource `status=""` label value, previously used to work around the same issue the nodes metric had, now fixed properly.


diegoakechi:
@ne-msft @steffenv-msft you might want to take a look at it, as it changes a bit the behavior of the cluster node metrics.

stefanotorresi:
marking as wip because it breaks current dashboards.

Issue #147, opened 2020-04-09
corosync (v3)  ring error - debian 10


kilida2020:
Hi,

I am not able to collect the metrics of the corosync ring, present the message below:

INFO[0000] Serving metrics on 0.0.0.0:9664
DEBU[0004] Collecting corosync metrics...
DEBU[0004] Collecting pacemaker metrics...
DEBU[0004] Collecting DRBD metrics...
WARN[0004] cannot parse ring status: corosync-cfgtool returned unexpected output: Printing link status.
Local node ID 1
LINK ID 0
        addr    = 128.1.0.86
        status  = OK

Thanks,

kilida2020:
Hi,

I can't collect the metrics for the corosync ring, please display the message below:

INFO [0000] Displaying metrics at 0.0.0.0:9664
DEBU [0004] Collecting corosync metrics ...
DEBU [0004] Collecting pacemaker metrics ...
DEBU [0004] Collecting DRBD metrics ...
WARN [0004] cannot analyze the status of the ring: corosync-cfgtool returned an unexpected exit: Printing link status.
Local node ID 1
LINK 0
address = 128.1.0.86
status = OK

Thanks

MalloZup:
@kilida2020 thx for issue. 

I have checked your issue. IT seems that you don't have any RING in your clusters, so that is why you get this warning.

can you run
```
 sudo corosync-cfgtool -s
```


you should have something like this 
```
 sudo corosync-cfgtool -s
Printing ring status.
Local node ID 33554442
RING ID 0
	id	= 10.0.0.1
	status	= ring 0 active with no faults```

If you have no ring this means you need to configure corosync. (check upstream docs)

Let me know

Best 
:sun_with_face:

kilida2020:
Hi,

This is my corosync configuration file.
You can analyze, I see that the RING is configured.

totem {
  version: 2
  cluster_name: debian
  secauth: off
  transport:udpu
  interface {
    ringnumber: 0
    bindnetaddr: 128.1.0.0
    broadcast: yes
    mcastport: 5405
  }
}

nodelist {
  node {
    ring0_addr: 128.1.0.86
    name: srvlabm
    nodeid: 1
  }
  node {
    ring0_addr: 128.1.0.87
    name: srvlabs
    nodeid: 2
  }
}

quorum {
  provider: corosync_votequorum
  two_node: 1
  wait_for_all: 1
  last_man_standing: 1
[...]

PR #156, opened 2020-05-27
support parsing corosync v2.99.0+ output

rmakram-ims:
This is to address this issue https://github.com/ClusterLabs/ha_cluster_exporter/issues/147.

### DIFF
--- a/collector/corosync/parser.go
+++ b/collector/corosync/parser.go
@@ -96,7 +96,11 @@ func parseRingIdAndSeq(cfgToolOutput []byte) (string, uint64, error) {
 	nodeRe := regexp.MustCompile(`(?m)Ring ID:\s+(\w+)/(\d+)`)
 	matches := nodeRe.FindSubmatch(cfgToolOutput)
 	if matches == nil {
-		return "", 0, errors.New("could not find Ring ID line")
+		nodeRe = regexp.MustCompile(`(?m)Ring ID:\s+(\w+).(\d+)`)
+		matches = nodeRe.FindSubmatch(cfgToolOutput)
+		if matches == nil {
+			return "", 0, errors.New("could not find Ring ID line")
+		}
 	}
 
 	seq, err := strconv.Atoi(string(matches[2]))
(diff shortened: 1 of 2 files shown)


collector/corosync/parser.go
```diff
@@ -96,7 +96,11 @@ func parseRingIdAndSeq(cfgToolOutput []byte) (string, uint64, error) {
 	nodeRe := regexp.MustCompile(`(?m)Ring ID:\s+(\w+)/(\d+)`)
 	matches := nodeRe.FindSubmatch(cfgToolOutput)
 	if matches == nil {
-		return "", 0, errors.New("could not find Ring ID line")
+		nodeRe = regexp.MustCompile(`(?m)Ring ID:\s+(\w+).(\d+)`)
```
stefanotorresi, line 99: since there is so little changing, maybe we could reuse the same regex instead of adding an additional if block?

`(?m)Ring ID:\s+(\w+).|/(\d+)` should do the trick.

collector/corosync/parser.go
```diff
@@ -128,8 +132,18 @@ func parseRings(cfgToolOutput []byte) []Ring {
...
+	// in corosync v2.99.0+ this has changed to
+	/*
+	   Link ID 0
+	   	addr	= 192.168.125.15
+	   	status	= ring 0 active with no faults
+	*/
 	re := regexp.MustCompile(`(?m)RING ID (?P<number>\d+)\s+id \s*= (?P<address>.+)\s+status \s*= (?P<status>.+)`)
 	matches := re.FindAllSubmatch(cfgToolOutput, -1)
+	if matches == nil {
+		re = regexp.MustCompile(`(?m)Link ID (?P<number>\d+)\s+addr \s*= (?P<address>.+)\s+status \s*= (?P<status>.+)`)
```
stefanotorresi, line 144: same as above

collector/corosync/parser.go
rmakram-ims: sure!

rmakram-ims: done

collector/corosync/parser.go
rmakram-ims: done

diegoakechi:
And I like even more when they write unit tests! =) Thanks a lot @rmakram-ims

rmakram-ims:
my pleasure!

rmakram-ims:
any chance of having a new release soon?

MalloZup:
@rmakram-ims  yes we can have a new release soonish thx for contrib

MalloZup:
as you see there is already the draft. I think we could do it tomorrow https://github.com/ClusterLabs/ha_cluster_exporter/releases

rmakram-ims:
awesome! thank you guys

stefanotorresi:
> I think we could do it tomorrow

more like next week, we need to implement a few new things first. 😉

rmakram-ims:
not a problem, I will build the binary from the current master branch then until you guys release the next version. great work by the way, this exporter is awesome!

PR #160, opened 2020-06-12
Grafana dashboard w/ OBS packaging

stefanotorresi:
This PR bundles the Grafana dashboards in this repository and introduce the OBS packaging to install those on top of Grafana.


packaging/obs/grafana-ha-cluster-dashboards/grafana-ha-cluster-dashboards.spec
```diff
@@ -0,0 +1,52 @@
...
+Version:        0
+Release:        0
+License:        Apache-2.0
+Summary:        Grafana Dashboards displaying metrics about a Pacemaker/Corosync High Availability Cluster.
+Group:          System/Monitoring
+Url:            https://github.com/ClusterLabs/ha_cluster_exporter
+Source:         %{name}-%{version}.tar.gz
+BuildArch:      noarch
+Requires:       grafana
+BuildRequires:  grafana
```
arbulu89, line 27: Is this required? I mean, to put some files in some folders during the build maybe grafana is not needed. If we can avoid it better hehe You know how the dependencies during the build time can be quite troubling

stefanotorresi: I initially tried to do without it, but we require the whole directory structure put in place by the Grafana package.
I guess I could duplicate some of the spec statements, but then what defines what package owns what path? Wouldn't we have conflicts?

dashboards/README.md
```diff
@@ -0,0 +1,3 @@
+# Grafana dashboards
+
+This directory contains the Grafana dashboards that leverage this exporter, along with a provider configuration file.
```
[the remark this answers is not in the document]
stefanotorresi: Yeah, this document is very much provisional, I even still have to decide if this should go in the project docs directory.

stefanotorresi:
> do we need an additional package? maybe we can just include this 3 files on the exporter package without having another package to maintain.
> What are the benifits of splitting it so small package?

we don't install the exporter and the dashboards in the same system.

stefanotorresi:
@arbulu89 

1. Yeah, the documentation surely needs improvements. I wanted to first experiment with the rpm, I will see to improve it at a later stage, along with the CI.
1. for the time being the versioning is handled manually and independently: we set the number in the `_service` file and then append just the commit timestamp and hash.
2. Yes, we require Grafana which is provided by SUMA. This package wouldn't have any sense without Grafana installed. As long as Grafana will be delivered by SUMA, this package will only be available to customers buying bundled licenses of both SUMA and SLEHA.
3. Tor the time being the versioning scheme is handled manually by changing the `_service` file. The only automated part is the `+git.%timestamp.%hash` suffix. Versioning will always be separate, as each release and commit regarding the exporter doesn't necessarily mean an update of the dashboards. This will be handled by reworking the CI a bit, in another PR.

dashboards/README.md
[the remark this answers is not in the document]
MalloZup: @stefanotorresi  I would reference the dashboard in our top-level `readme`.

To install a grafana dashboard is pretty straightforward imho, we just say import the dashboard. and we can reference upstream.

But we need to make it visible on the top-level readme so an user know there is a dashboard

Makefile
```diff
@@ -53,29 +52,41 @@ coverage:
...
-
-clean-obs:
-	rm -rf build/obs
-
-obs-workdir: clean-obs
-	@mkdir -p build/obs
-	osc checkout $(OBS_PROJECT) $(OBS_PACKAGE) -o build/obs
-	rm -f build/obs/*.tar.gz
-	cp -rv packaging/obs/* build/obs/
+exporter-obs-workdir:
```
MalloZup, line 59: what is used for the `obs-workdir`?

packaging/obs/grafana-ha-cluster-dashboards/grafana-ha-cluster-dashboards.spec
MalloZup: I think is ok to require grafana, `BuildRequires` maybe we can spare it

Makefile
stefanotorresi: hmm?

stefanotorresi:
> do we need the screenshot.png? it might be depracted dashboards/screenshot.png

yes, the plan is to show what the dashboard looks like in the markdown docs.

dashboards/README.md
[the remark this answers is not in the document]
stefanotorresi: yep, I'm working on the docs right now!

packaging/obs/grafana-ha-cluster-dashboards/grafana-ha-cluster-dashboards.spec
stefanotorresi: the build will fail without that `BuildRequire` with the following error:
```
[   28s] grafana-ha-cluster-dashboards-1.0.0+git.1591970022.e33fbe9-1.1.noarch.rpm: directories not owned by a package:
[   28s]  - /etc/grafana
[   28s]  - /etc/grafana/provisioning
[   28s]  - /etc/grafana/provisioning/dashboards
[   28s]  - /var/lib/grafana
[   28s]  - /var/lib/grafana/dashboards
```

Makefile
MalloZup: I just forget the term

PR #163, opened 2020-06-17
GitHub actions workflows

stefanotorresi:
This PR removes Travis as a CI system and introduces GitHub actions via two separated worfklows for the exporter and the dashboards.

The following GitHub secrets need to be configured to make this work:
- `OBS_DOWNSTREAM_PROJECT`
- `OBS_PASS`
- `OBS_PROJECT`
- `OBS_USER`


.github/workflows/dashboards-ci.yml
```diff
@@ -0,0 +1,49 @@
...
+    - uses: actions/setup-node@v1
+    - run: npm -g install jsonlint
+    - name: validate JSON
+      run: find dashboards -name "*.json" -type f -exec jsonlint -c {} \;
+
+  obs-commit:
+    needs: build
+    if: github.ref == 'refs/heads/master'
+    runs-on: ubuntu-latest
+    container: shap/continuous_deliver
```
MalloZup, line 31: is this a container we have ? maybe we need to update confluence page regarding this

.github/workflows/exporter-ci.yml
```diff
@@ -0,0 +1,103 @@
+name: Exporter CI
```
MalloZup, line 1: for all workflow, I would like to have an higher lever description to see what is the intent of the workflow

like. Build and test XX foo .

.github/workflows/dashboards-ci.yml
```diff
@@ -0,0 +1,49 @@
+name: Dashboards CI
```
MalloZup, line 1: same here, highlevel rationale is alwasy appreciate as comment

check json on pr, build only on master

.github/workflows/dashboards-ci.yml
stefanotorresi: yeah this is hosted on docker hub and sourced from https://github.com/arbulu89/continuous-delivery

.github/workflows/exporter-ci.yml
stefanotorresi: I will add some comments with a more exhaustive description: I can't use the `name` field, that's meant to be a short name.

MalloZup: yeah just add comment

.github/workflows/dashboards-ci.yml
MalloZup: ok I think this was an `howto` tutorial. I will move it to `CI` session. I think personally the howto section is somehow to much generic like `utils` directory without any context

MalloZup: @stefanotorresi  i'm moving the page on the CI section. feel free to add it maybe with github action context too

stefanotorresi: yeah, good call!

PR #172, opened 2020-08-05
Add qdevice membership parsing and metric label

### LINKED ISSUE #170: corosync membership parsing with qdevice enabled
Same problem as in #154 but with corosync-2.4.5-4.el7.x86_64 on CentOS 7

Jul 16 13:01:57 nfs01 ha_cluster_exporter: time="2020-07-16T13:01:57+03:00" level=warning msg="'corosync' collector scrape failed: corosync parser error: could not parse members in corosync-quorumtool output: could not find membership information"

The output of corosync-quorumtool now looks like:

```
Quorum information
------------------
Date:             Thu Jul 16 12:40:44 2020
Quorum provider:  corosync_votequorum
Nodes:            2
Node ID:          1
Ring ID:          1/139
Quorate:          Yes

Votequorum information
----------------------
Expected votes:   3
Highest expected: 3
Total votes:      3
Quorum:           2  
Flags:            Quorate Qdevice 

Membership information
----------------------
    Nodeid      Votes    Qdevice Name
         1          1    A,V,NMW nfs01 (local)
         2          1    A,V,NMW nfs02
         0          1            Qdevice
```

See 'Qdevice Name' not 'Name' in column titles under Membership information.

MalloZup:
Fix #170 

There is only one special case, we might need to handle:

```ha_cluster_corosync_member_votes{local="false",node="Qdevice",node_id="0",qdevice=""} 1```

when the qdevice is active

```
Quorum information
------------------
Date:             Thu Jul 16 12:40:44 2020
Quorum provider:  corosync_votequorum
Nodes:            2
Node ID:          1
Ring ID:          1/139
Quorate:          Yes

Votequorum information
----------------------
Expected votes:   3
Highest expected: 3
Total votes:      3
Quorum:           2  
Flags:            Quorate Qdevice 

Membership information
----------------------
    Nodeid      Votes    Qdevice Name
         1          1    A,V,NMW nfs01 (local)
         2          1    A,V,NMW nfs02
         0          1            Qdevice
```

the table qdevice is empty so we create an empty metric.

I was thinking in order to avoid an empty metric we should populated the label `qdevice` with `qdevice` for indicating that it is the qdevice.

@gao-yan @nick-wang let me know what do you think

### DIFF
--- a/collector/corosync/parser.go
+++ b/collector/corosync/parser.go
@@ -227,7 +227,7 @@ func parseMembers(quorumToolOutput []byte) (members []Member, err error) {
 	/*
 		1          1  A,V,NMW 192.168.125.24 (local)
 	*/
-	linesRE := regexp.MustCompile(`(?m)(?P<node_id>\w+)\s+(?P<votes>\d+)\s+(?P<qdevice>(\w,?)+)\s+(?P<name>[\w-\.]+)(?:\s(?P<local>\(local\)))?\n?`)
+	linesRE := regexp.MustCompile(`(?m)(?P<node_id>\w+)\s+(?P<votes>\d+)\s+(?P<qdevice>(\w,?)+)?\s+(?P<name>[\w-\.]+)(?:\s(?P<local>\(local\)))?\n?`)
 	linesMatches := linesRE.FindAllSubmatch(sectionMatch[1], -1)
 	for _, match := range linesMatches {
 		matches := extractRENamedCaptureGroups(linesRE, match)
--- a/collector/corosync/parser.go
+++ b/collector/corosync/parser.go
@@ -35,10 +35,11 @@ type Ring struct {
 }
 
 type Member struct {
-	Id    string
-	Name  string
-	Votes uint64
-	Local bool
+	Id      string
+	Name    string
+	Qdevice string
+	Votes   uint64
+	Local   bool
 }
 
 func NewParser() Parser {
@@ -211,11 +212,12 @@ func parseMembers(quorumToolOutput []byte) (members []Member, err error) {
 	/*
 	   Membership information
 	   ----------------------
-	      Nodeid      Votes Name
-	   		1          1 192.168.125.24
-	   		2          1 192.168.125.25 (local)
+	   Nodeid      Votes    Qdevice Name
+	        1          1    A,V,NMW nfs01 (local)
+	        2          1    A,V,NMW nfs02
+	        0          1            Qdevice
 	*/
-	sectionRE := regexp.MustCompile(`(?m)Membership information\n-+\s+Nodeid\s+Votes\s+Name\n+((?:.*\n?)+)`)
+	sectionRE := regexp.MustCompile(`(?m)Membership information\n-+\s+Nodeid\s+Votes\s+Qdevice\s+Name\n+((?:.*\n?)+)`)
 	sectionMatch := sectionRE.FindSubmatch(quorumToolOutput)
 	if sectionMatch == nil {
 		return nil, errors.New("could not find membership information")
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by stefanotorresi** [comment]
> I was thinking in order to avoid an empty metric we should populated the label `qdevice` with `qdevice` for indicating that it is the qdevice.

Well, it's not really empty, you can see it's the Qdevice from the name! I don't think this is that special, after all.

File: collector/corosync/parser.go [hunk 1]
```diff
@@ -211,21 +212,22 @@ func parseMembers(quorumToolOutput []byte) (members []Member, err error) {
 	/*
 	   Membership information
 	   ----------------------
-	      Nodeid      Votes Name
-	   		1          1 192.168.125.24
-	   		2          1 192.168.125.25 (local)
+	   Nodeid      Votes    Qdevice Name
+	        1          1    A,V,NMW nfs01 (local)
+	        2          1    A,V,NMW nfs02
```
stefanotorresi, line 217: do these values change at runtime?

MalloZup: yes the can change https://susedoc.github.io/doc-sleha/master/single-html/SLE-HA-guide/#co-quorum-nodes

stefanotorresi: hmm, that's what I thought. We shouldn't put them in a label then: I would suggest to not include these values at all at the moment. If we are asked to export them, we'll need to introduce new metrics with static labels.

MalloZup: why we shouldn't put them in a label? Can  you elaborate this point why do you think we shouldn't?

stefanotorresi: I there is one key takeaway to be remembered from Promcon, is that most Prometheus users are using labels the wrong way, and we are doing this too, a lot.
Labels are not intended to be used for values that change over the lifecycle of what is meant to be a single timeseries entry.
Prometheus timeseries DB is effectively an indexed key-value store, and labels are part of the key. Every new label value will create a new timeseries.

In this specific case, for example, when the communication between the node and the qdevice changes status, we would end up with two timeseries for the same node.

A metric line exposing nominal status would be like this:
```
ha_cluster_corosync_member_votes{local="false",node="node01",node_id="0",qdevice="A,V,NMW"} 1
```

While a metric exposing a problem would be like this:
```
ha_cluster_corosync_member_votes{local="false",node="node01",node_id="0",qdevice="NA,NV,NMW"} 1
```

The thing that changes is just the `qdevice` label, not **the value of the metric**. These are effectively two different timeseries, they're like two separate entities, instead of being 2 different statuses of the same entity.

What these metrics are meant to represent is just the number of votes, not the status of the communication between the nodes and the qdevice. It doesn't make much sense to count the vote of a node differently when it's in a status rather than another, but this is exactly what happens if we put values that change at runtime in the labels.

This causes _a lot_ of problems when consuming metrics, and in the long term puts a lot of weight in the DB index, eventually causing serious performance problems down the line.

This by far the most common and easiest mistake to make, we are already affected by this in multiple cases, so we need to avoid this in the future, and hopefully fix all the old metrics for v2 of the exporter.

A way to correctly represent the qdevice status of nodes would be a completely new metric dedicated to this, where the status is in a label, but _the label doesn't change over time_, its _value_ does.

```
ha_cluster_corosync_qdevice_status{node="node01",status="NA"} 0
ha_cluster_corosync_qdevice_status{node="node01",status="A"} 0
ha_cluster_corosync_qdevice_status{node="node01",status="V"} 1
ha_cluster_corosync_qdevice_status{node="node01",status="NV"} 0
ha_cluster_corosync_qdevice_status{node="node01",status="NMW"} 1
ha_cluster_corosync_qdevice_status{node="node01",status="MW"} 0
```

This way we'll have a _static_ number of timeseries for a single entity (the node), that _will **not increase** over time_. The only thing changing over time is _their value_.

MalloZup: Ok fair point. Regarding the prometheus performance , we don't have any real data and this are just assumptions in the context of our small exporter.

But in sake of the meaning of the metric  make sense.  I will adapt the pr accordingly thx

stefanotorresi: > we don't have any real data and this are just assumptions in the context of our small exporter.

Hmm, just to clarify: these are not my assumptions. Putting dynamic content in labels is a practice abundantly advertised against within the Prometheus community and was a very prominent topic during the last PromCon.

No, I don't have hard data to show, so yes, this is anecdotal, but a lot of people more knowledgeable than me keep saying to not do this, so I think we should listen to them. :wink: 

To put it simply: we have been doing it very wrong all the time; the `handb_exporter`, which puts entire dynamically generated error messages inside labels, is a ticking bomb waiting to explode, and it will cause production workloads to eventually malfunction.

stefanotorresi:
thanks Dario!

MalloZup:
thx

collector/corosync/parser.go
MalloZup: I agree with this, but I guess the performance impact is not critical in our exporter . although I agree that we shouldn't overload the label and follow upstream recommendations when they make sense . that is fair point

PR #174, opened 2020-09-01
Implement SBD watchdog and msgwait metrics

MalloZup:
# todo:
- [x] add documentation for metrics

# Note:

I have implemented the metric  like 

```
ha_cluster_sbd_devices{device="/dev/vdd",status="healthy"} 1
ha_cluster_sbd_devices_watchdog_timeout{device="/dev/vdc"} 60
ha_cluster_sbd_devices_msgwait_timeout{device="/dev/vdd"} 120
```

I didn't found any useful doc about the `loop` and  `allocate` metric so I even wondered if we should expose them. (In fact I haven't but if they are meant to be useful, we should definitely document this in the suse-doc)
See:
https://documentation.suse.com/sle-ha/15-SP2/html/SLE-HA-all/cha-ha-storage-protect.html#sec-ha-storage-protect-watchdog-timings

@diegoakechi cc
@nick-wang cc

I personally think that if we are exposing only 2 metrics we can create for each one a new metric without label.

If we want to expose 4 then I will refactor accordingly with a `type` label. I am ok with both directions

### DIFF
--- a/collector/sbd/sbd.go
+++ b/collector/sbd/sbd.go
@@ -35,7 +35,7 @@ func NewCollector(sbdPath string, sbdConfigPath string) (*sbdCollector, error) {
 	}
 
 	c.SetDescriptor("devices", "SBD devices; one line per device", []string{"device", "status"})
-	c.SetDescriptor("sbd_timeout", "sbd watchdog timeout", []string{"device", "type"})
+	c.SetDescriptor("timeouts", "SBD timeouts for each device and type", []string{"device", "type"})
 
 	return c, nil
 }
@@ -73,11 +73,11 @@ func (c *sbdCollector) CollectWithError(ch chan<- prometheus.Metric) error {
 
 	sbdWatchdogs, sbdMsgWaits := c.getSbdTimeouts(sbdDevices)
 	for sbdDev, sbdWatchdog := range sbdWatchdogs {
-		ch <- c.MakeGaugeMetric("sbd_timeout", sbdWatchdog, sbdDev, "watchdog_timeout")
+		ch <- c.MakeGaugeMetric("timeouts", sbdWatchdog, sbdDev, "watchdog")
 	}
 
 	for sbdDev, sbdMsgWait := range sbdMsgWaits {
-		ch <- c.MakeGaugeMetric("sbd_timeout", sbdMsgWait, sbdDev, "msgwait_timeout")
+		ch <- c.MakeGaugeMetric("timeouts", sbdMsgWait, sbdDev, "msgwait")
 	}
 
 	return nil
--- a/collector/sbd/sbd.go
+++ b/collector/sbd/sbd.go
@@ -35,8 +35,7 @@ func NewCollector(sbdPath string, sbdConfigPath string) (*sbdCollector, error) {
 	}
 
 	c.SetDescriptor("devices", "SBD devices; one line per device", []string{"device", "status"})
-	c.SetDescriptor("watchdog_timeout", "sbd watchdog timeout", []string{"device"})
-	c.SetDescriptor("msgwait_timeout", "sbd msgwait timeout", []string{"device"})
+	c.SetDescriptor("sbd_timeout", "sbd watchdog timeout", []string{"device", "type"})
 
 	return c, nil
 }
@@ -72,14 +71,13 @@ func (c *sbdCollector) CollectWithError(ch chan<- prometheus.Metric) error {
 		ch <- c.MakeGaugeMetric("devices", 1, sbdDev, sbdStatus)
 	}
 
-	sbdWatchdogs := c.getSbdWatchDogTimeout(sbdDevices)
+	sbdWatchdogs, sbdMsgWaits := c.getSbdTimeouts(sbdDevices)
 	for sbdDev, sbdWatchdog := range sbdWatchdogs {
-		ch <- c.MakeGaugeMetric("watchdog_timeout", sbdWatchdog, sbdDev)
+		ch <- c.MakeGaugeMetric("sbd_timeout", sbdWatchdog, sbdDev, "watchdog_timeout")
 	}
 
-	sbdMsgWaits := c.getSbdMsgWaitTimeout(sbdDevices)
 	for sbdDev, sbdMsgWait := range sbdMsgWaits {
-		ch <- c.MakeGaugeMetric("msgwait_timeout", sbdMsgWait, sbdDev)
+		ch <- c.MakeGaugeMetric("sbd_timeout", sbdMsgWait, sbdDev, "msgwait_timeout")
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


diegoakechi:
@MalloZup How are these kind of metric being formatted on other cases or maybe on other exporters? I think the usage of label more flexible and elegant, but at the same time it might make harder for the consumer. Imagine that we use label and implement only 2 metrics. When a third label is introduced we might break dashboards that might not have consistent filters, while a new metric would be ignored. 

With that my vote is for separated metrics.

stefanotorresi (changes requested):
@MalloZup as I mentioned in the JIRA issues, the way I would go about this is to add a single metric with both a `device` and a `ŧype` label, and add all the timeouts.
There is no real downside to expose also the other two types that you have left out.

The final result should be something like this:

```
# TYPE ha_cluster_sbd_device_timeout gauge
ha_cluster_sbd_device_timeout{type="watchdog",device="/dev/vdc"} 5
ha_cluster_sbd_device_timeout{type="allocate",device="/dev/vdc"} 2
ha_cluster_sbd_device_timeout{type="loop",device="/dev/vdc"} 1
ha_cluster_sbd_device_timeout{type="msgwait",device="/dev/vdc"} 10
```

stefanotorresi:
@diegoakechi for these kind of use cases, the best practice is exactly the opposite of using separate metrics. See https://prometheus.io/docs/practices/instrumentation/#use-labels

collector/sbd/sbd.go
```diff
@@ -132,3 +146,57 @@ func (c *sbdCollector) getSbdDeviceStatuses(sbdDevices []string) map[string]stri
...
+		if msgWaitTimeout == nil {
+			continue
+		}
+		// map the timeout to the device
+		if s, err := strconv.ParseFloat(msgWaitTimeout[0], 64); err == nil {
+			sbdMsgWaits[sbdDev] = s
+		}
+	}
+	return sbdMsgWaits
+}
```
stefanotorresi, line 202: There is a lot of repetition here, maybe we could refactor these two functions into a single one that gets all the timeouts.
The returned map would be something like `map[string]map[string]float64`, where the first key is the device and the second is the type of the timeout.

diegoakechi:
> @diegoakechi for these kind of use cases, the best practice is exactly the opposite of using separate metrics. See https://prometheus.io/docs/practices/instrumentation/#use-labels

Sure, in that case lets go this way. Also, make sure to include all the metrics, so we avoid any filtering problem in the future. Thanks @stefanotorresi

collector/sbd/sbd.go
MalloZup: yes for a 1st version I wanted to have something done, and have a high-level pov before solving the details. I will refactor accordingly now

MalloZup: yes for a 1st version I wanted to have something done, and have a high-level pov before solving the details. I will refactor accordingly now

MalloZup:
@stefanotorresi  I'm ok to use the `type` even for 2 metric.

```
ha_cluster_sbd_device_timeout{type="watchdog",device="/dev/vdc" 1
ha_cluster_sbd_device_timeout{type="msgwait",device="/dev/vdc"} 1
```
as follow-up with @yan-gao discussion we don't need to expose the other 2 metrics since they are more confusing then helping


thx for reviews.

stefanotorresi:
> as follow-up with @yan-gao discussion we don't need to expose the other 2 metrics since they are more confusing then helping

okay, sounds good to me!

collector/sbd/sbd.go
```diff
@@ -35,8 +35,7 @@ func NewCollector(sbdPath string, sbdConfigPath string) (*sbdCollector, error) {
 	}
 
 	c.SetDescriptor("devices", "SBD devices; one line per device", []string{"device", "status"})
-	c.SetDescriptor("watchdog_timeout", "sbd watchdog timeout", []string{"device"})
-	c.SetDescriptor("msgwait_timeout", "sbd msgwait timeout", []string{"device"})
+	c.SetDescriptor("sbd_timeouts", "sbd watchdog timeout", []string{"device", "type"})
```
stefanotorresi, line 38: The description is a bit misleading, as we have both watchdog and msgwait timeouts, not just the watchdog one. Also we dont need the `sbd_` prefix, that's already added by the constructor.
Here is something more generic:
```suggestion
	c.SetDescriptor("timeouts", "SBD timeouts for each device and type", []string{"device", "type"})
```

collector/sbd/sbd.go
```diff
@@ -72,14 +71,13 @@ func (c *sbdCollector) CollectWithError(ch chan<- prometheus.Metric) error {
 		ch <- c.MakeGaugeMetric("devices", 1, sbdDev, sbdStatus)
 	}
 
-	sbdWatchdogs := c.getSbdWatchDogTimeout(sbdDevices)
+	sbdWatchdogs, sbdMsgWaits := c.getSbdTimeouts(sbdDevices)
 	for sbdDev, sbdWatchdog := range sbdWatchdogs {
-		ch <- c.MakeGaugeMetric("watchdog_timeout", sbdWatchdog, sbdDev)
+		ch <- c.MakeGaugeMetric("sbd_timeouts", sbdWatchdog, sbdDev, "watchdog_timeout")
```
stefanotorresi, line 76: I think we can avoid these repetitions:
```suggestion
		ch <- c.MakeGaugeMetric("timeouts", sbdWatchdog, sbdDev, "watchdog")
```

collector/sbd/sbd.go
```diff
@@ -72,14 +71,13 @@ func (c *sbdCollector) CollectWithError(ch chan<- prometheus.Metric) error {
...
 	}
 
-	sbdMsgWaits := c.getSbdMsgWaitTimeout(sbdDevices)
 	for sbdDev, sbdMsgWait := range sbdMsgWaits {
-		ch <- c.MakeGaugeMetric("msgwait_timeout", sbdMsgWait, sbdDev)
+		ch <- c.MakeGaugeMetric("sbd_timeouts", sbdMsgWait, sbdDev, "msgwait_timeout")
```
stefanotorresi, line 80: Same as above
```suggestion
		ch <- c.MakeGaugeMetric("timeouts", sbdMsgWait, sbdDev, "msgwait")
```

doc/metrics.md
```diff
@@ -214,6 +215,18 @@ Either the value is `1`, or the line is absent altogether.
...
+
+#### Description
+
+The SBD timeouts pro SBD device
+Value is an integer expessing the timeout
+
+#### Labels
+
+- `device`: the path of the SBD device
+- `type`:  either `watchdog_timeout` or `msgwait_timeout`
```
stefanotorresi, line 228:
```suggestion
- `type`:  either `watchdog` or `msgwait`
```

test/sbd.metrics
```diff
@@ -1,12 +1,10 @@
...
+# HELP ha_cluster_sbd_devices SBD devices; one line per device
+# TYPE ha_cluster_sbd_devices gauge
+ha_cluster_sbd_devices{device="/dev/vdc",status="healthy"} 1
+ha_cluster_sbd_devices{device="/dev/vdd",status="healthy"} 1
+# HELP ha_cluster_sbd_sbd_timeouts sbd watchdog timeout
+# TYPE ha_cluster_sbd_sbd_timeouts gauge
+ha_cluster_sbd_sbd_timeouts{device="/dev/vdc",type="msgwait_timeout"} 10
+ha_cluster_sbd_sbd_timeouts{device="/dev/vdc",type="watchdog_timeout"} 9
+ha_cluster_sbd_sbd_timeouts{device="/dev/vdd",type="msgwait_timeout"} 10
+ha_cluster_sbd_sbd_timeouts{device="/dev/vdd",type="watchdog_timeout"} 9
```
stefanotorresi, line 10:
```suggestion
ha_cluster_sbd_timeouts{device="/dev/vdc",type="msgwait"} 10
ha_cluster_sbd_timeouts{device="/dev/vdc",type="watchdog"} 9
ha_cluster_sbd_timeouts{device="/dev/vdd",type="msgwait"} 10
ha_cluster_sbd_timeouts{device="/dev/vdd",type="watchdog"} 9
```

test/sbd.metrics
```diff
@@ -1,12 +1,10 @@
...
- ha_cluster_sbd_msgwait_timeout{device="/dev/vdd"} 10
- # HELP ha_cluster_sbd_watchdog_timeout sbd watchdog timeout
- # TYPE ha_cluster_sbd_watchdog_timeout gauge
- ha_cluster_sbd_watchdog_timeout{device="/dev/vdc"} 9
- ha_cluster_sbd_watchdog_timeout{device="/dev/vdd"} 9
+# HELP ha_cluster_sbd_devices SBD devices; one line per device
+# TYPE ha_cluster_sbd_devices gauge
+ha_cluster_sbd_devices{device="/dev/vdc",status="healthy"} 1
+ha_cluster_sbd_devices{device="/dev/vdd",status="healthy"} 1
+# HELP ha_cluster_sbd_sbd_timeouts sbd watchdog timeout
```
stefanotorresi, line 5:
```suggestion
# HELP ha_cluster_sbd_timeouts SBD timeouts for each device and type
```

stefanotorresi (changes requested):
Looks good! I suggest just a few minor tweaks to the labels and names.

MalloZup (commented):
thx should be ok now

stefanotorresi (approved):

PR #180, opened 2020-12-02
remove explicit Pacemaker dependency from the systemd unit

stefanotorresi:
We don't actually need the Pacemaker service to be running, and we don't want to interact with it in case the system is rebooted (e.g. the Pacemaker service was temporarily disabled to do maintenance).


MalloZup (approved):
thx

PR #192, opened 2021-05-03
Update dependencies to latest releases

dirkmueller:
Also require go 1.14, which is the oldest still maintained
golang version (and some of the dependencies require at
least 1.14)


stefanotorresi (approved):
Thank you!

stefanotorresi (changes requested):
Actually, could you please update https://github.com/ClusterLabs/ha_cluster_exporter/blob/master/packaging/obs/prometheus-ha_cluster_exporter/prometheus-ha_cluster_exporter.spec accordingly?

dirkmueller:
@stefanotorresi done

stefanotorresi (approved):

PR #250, opened 2024-03-20
add supportconfig plugin (jsc#PED-2560)

angelabriel:
during a support training there was the idea of having a supportconfig  plugin for the exporter packages to help with problem analysis

### DIFF
--- a/supportconfig-ha_cluster_exporter
+++ b/supportconfig-ha_cluster_exporter
@@ -1,0 +1,87 @@
+#!/bin/bash
+set -u
+
+# supportconfig plugin for ha_cluster_exporter
+#
+# v1.0
+#
+# February 2024    v1.0    first release
+
+SVER='1.0.0'
+TITLE="SUSE supportconfig plugin for ha_cluster_exporter"
+
+function display_package_info() {
+    echo -e "\n#==[ Command ]======================================#"
+    echo -e "# rpm -qi ${1}"
+    rpm -qi "${1}"
+
+    echo -e "\n#==[ Command ]======================================#"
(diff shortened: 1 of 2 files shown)


supportconfig-ha_cluster_exporter
```diff
@@ -0,0 +1,87 @@
...
+for file in /usr/etc/ha_cluster_exporter.{yaml,json,toml} /etc/ha_cluster_exporter.{yaml,json,toml} /usr/etc/ha_cluster_exporter.web.yaml /etc/ha_cluster_exporter.web.yaml; do
+    [ -e "${file}" ] && { display_file_stat "${file}" ; display_file "${file}" ; echo ; }
+done
+
+display_file_stat /etc/sysconfig/prometheus-ha_cluster_exporter
+display_file /etc/sysconfig/prometheus-ha_cluster_exporter
+
+#log infos from system log
+display_cmd "grep -E -i 'ha_cluster_exporter\[.*\]:' /var/log/messages"
+display_cmd "ss -tulpan | grep exporter"
```
stefanotorresi, line 84: hey Angela, I saw you did put this very same line in all three exporters, isn't the grep a little too generic? this would match anything called "exporter". Shouldn't we grep for the exact name?

angelabriel: Hi Stefano, the idea behind the generic grep was to see, if other exporters are available too and may interfere or affect our exporters.

stefanotorresi: ah okay, makes sense, thanks for clarifying.

PR #260, opened 2024-07-24
add support to show overall cluster maintenance mode

scrusanhrt:
if you put an entire cluster into maintenance mode (for whatever reason, sometimes valid) via "crm configure property maintenance-mode=true", then it won't be shown by this exporter.

this is because it currently only checks node properties. If you set an individual node to maintenance-mode, then it *will* show up.

new functionality:

```$ curl -o - <host>:9664/metrics 2>/dev/null| grep maint
# HELP ha_cluster_pacemaker_maintenance_mode_enabled Whether or not cluster wide maintenance-mode is enabled
# TYPE ha_cluster_pacemaker_maintenance_mode_enabled gauge
ha_cluster_pacemaker_maintenance_mode_enabled 1
ha_cluster_pacemaker_nodes{node="<host>",status="maintenance",type="member"} 0
```

tests pass via "make test"


stefanotorresi:
Cool! Thanks a lot for your contribution!

PR #284, opened 2026-02-19
add support for exporting transient attributes from cib

juliuswaldmann:
This PR adds the exporting of transient attributes form the cib


juliuswaldmann:
@stefanotorresi what do you think? ^^

